PR #12, opened 2020-03-19
Update test workspace, add fixtures

xbrianh:
Use Terra workspace `terra-notebook-utils-tests` as test bed.

### DIFF
--- a/tests/fixtures/populate_workspace_data.py
+++ b/tests/fixtures/populate_workspace_data.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+"""
+Upload fixtures into Terra data model
+"""
+import os
+import sys
+import json
+
+from firecloud import fiss
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', ".."))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+import tests.config
+WORKSPACE_GOOGLE_PROJECT = os.environ['GOOGLE_PROJECT']
+WORKSPACE_NAME = os.environ['WORKSPACE_NAME']
+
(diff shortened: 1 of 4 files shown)


tests/test_terra_notebook_utils.py
```diff
@@ -54,11 +58,6 @@ def test_oneshot_copy(self):
     def test_multipart_copy(self):
         drs.copy(self.drs_url, "test_oneshot_object", multipart_threshold=1024 * 1024)
 
-    # Probably don't want to run this test very often. Once a week?
```
DailyDreaming: Should this be kept separately and manually run?

tests/fixtures/populate_workspace_data.py
```diff
@@ -0,0 +1,49 @@
...
+    resp.raise_for_status()
+    for ent in resp.json():
+        yield ent
+
+def delete_entities(entities):
+    resp = fiss.fapi.delete_entities(WORKSPACE_GOOGLE_PROJECT, WORKSPACE_NAME, entities)
+    resp.raise_for_status()
+
+def delete_all_entities():
+    entities_to_delete = [dict(entityType=e['entityType'], entityName=e['name'])
```
DailyDreaming: Should this yield each element to a delete single item function instead of building the entire dictionary at once?

xbrianh: This deletes the entire contents of the workspace, so I think it's OK for now.

tests/test_terra_notebook_utils.py
xbrianh: Perhaps we can use our swanky new CI/CD system to execute this once a week, or once a month.

PR #13, opened 2020-03-19
Refactor progress indicators

xbrianh:
- Add progress reporter
- Rate limit output to keep things perty
- Add tests


terra_notebook_utils/progress.py
```diff
@@ -0,0 +1,126 @@
...
+        self.steps_completed += number_of_steps
+        slots_filled = int(self._length * self.steps_completed / self.number_of_steps)
+        duration = ceil(time.time() - self._start_time)
+        portion_complete = self.steps_completed / self.number_of_steps
+        bar = "#" * slots_filled
+        bar += " " * (self._length - slots_filled)
+        bar += f" {floor(100 * portion_complete)}%"
+        if self.size is not None:
+            bar += f" of {self.size}{self.units}"
+            bar += f" %.2f{self.units}/s" % (self.size * portion_complete / duration)
```
DailyDreaming: Nit: Mixing f-strings and %-formatting.  XD

terra_notebook_utils/progress.py
```diff
@@ -0,0 +1,126 @@
...
+        self._last_call = time.time() - 10 * self.period
+
+    def __call__(self, func):
+        def wrapper(*args, **kwargs):
+            now = time.time()
+            if now - self._last_call >= self.period:
+                self._last_call = now
+                return func(*args, **kwargs)
+            elif self.exception is not None:
+                raise self.exception("Too soon")
```
DailyDreaming: lol

terra_notebook_utils/progress.py
```diff
@@ -0,0 +1,126 @@
...
+    def checkpoint(self, _units_processed: int):
+        self._units_processed = [self._units_processed[-1], self._units_processed[-1] + _units_processed]
+        self._checkpoint_time = [self._checkpoint_time[-1], time.time()]
+        self._print()
+
+    @RateLimited(2.0)
+    def _print(self):
+        with self._lock:
+            print("\r",
+                  f"%9i {self.units} processed," % self._units_processed[-1],
```
DailyDreaming: Same formatting nit here and below.

terra_notebook_utils/progress.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class ProgressBar(AbstractContextManager):
+    def __init__(self, number_of_steps, prefix="", units="", size: int=None):
+        self.number_of_steps = number_of_steps
+        self.steps_completed = 0
+        self.prefix = prefix
+        self.units = units
+        self.size = size
+        self._length = 40
```
DailyDreaming: Seems like a magic number?  Maybe a comment?

terra_notebook_utils/progress.py
```diff
@@ -0,0 +1,126 @@
...
+    @property
+    def duration(self):
+        return self._checkpoint_time[-1] - self.start_time
+
+    @property
+    def average_rate(self):
+        return self._units_processed[-1] / self.duration
+
+    @property
+    def instantaneous_rate(self):
```
DailyDreaming: Nit: is this function called?

terra_notebook_utils/progress.py
```diff
@@ -0,0 +1,126 @@
...
+    Decorator to rate limit method calls. Raises `exception` if method is called more quickly than
+    `rate` times per second, or does nothing if exception is None.
+    """
+    def __init__(self, rate: float, exception: Exception=None):
+        self.period = 1 / rate
+        self.exception = exception
+        self.reset()
+
+    def reset(self):
+        self._last_call = time.time() - 10 * self.period
```
DailyDreaming: What does the `10` account for?

tests/test_terra_notebook_utils.py
```diff
@@ -86,5 +90,49 @@ def test_extract(self):
...
+            @progress.RateLimited(rate, MockRateLimitedException)
+            def raising_rate_limited_func():
+                pass
+
+            with self.assertRaises(MockRateLimitedException):
+                raising_rate_limited_func()
+                raising_rate_limited_func()
+
+        with self.subTest("Should be able to avoid exceptions using reset"):
+            raising_rate_limited_func.reset()
```
DailyDreaming: Is it consistent if this is done in a `for loop` for many more times?

tests/test_terra_notebook_utils.py
```diff
@@ -86,5 +90,49 @@ def test_extract(self):
...
+        with self.subTest("Should be rate limited to expected number of calls"):
+            call_info = dict(number_of_calls=0)
+
+            @progress.RateLimited(rate)
+            def rate_limited_func():
+                call_info['number_of_calls'] += 1
+
+            start_time = time.time()
+            duration = 0
+            while duration <= 4.5 * 1 / rate:
```
DailyDreaming: I'd expect this to be `5 * 1`?

tests/test_terra_notebook_utils.py
```diff
@@ -86,5 +90,49 @@ def test_extract(self):
...
+        with self.subTest("Should get expected exception for calling too quickly"):
+            class MockRateLimitedException(Exception):
+                pass
+
+            @progress.RateLimited(rate, MockRateLimitedException)
+            def raising_rate_limited_func():
+                pass
+
+            with self.assertRaises(MockRateLimitedException):
+                raising_rate_limited_func()
```
DailyDreaming: Why does this call `raising_rate_limited_func()` twice?

terra_notebook_utils/progress.py
xbrianh: It's just an arbitrary factor to guarantee the function is not rate limited on the next call. anything larger than 1 should work.

terra_notebook_utils/progress.py
xbrianh: Not magic so much as a choice for a reasonable length. I'll add a comment.

terra_notebook_utils/progress.py
xbrianh: What's wrong with it?

terra_notebook_utils/progress.py
xbrianh: Nope, but it seems like it'll be occasionally useful.

tests/test_terra_notebook_utils.py
xbrianh: The first call is at `t=0`. If we waited for 5 seconds to ellapse I'd expect to get 6 calls, not 5.

tests/test_terra_notebook_utils.py
xbrianh: Calling it once wouldn't cause the rate limiting logic to trigger.

tests/test_terra_notebook_utils.py
xbrianh: That would work, but wouldn't test any additional logic.

Switched to loops to improve aesthetics.

PR #17, opened 2020-04-06
Scripts

xbrianh:


scripts/file-finder.py
```diff
@@ -0,0 +1,52 @@
...
+from datetime import datetime
+import json
+from firecloud import fiss
+from firecloud.errors import FireCloudServerError
+import firecloud.api as fapi
+import numpy as np
+import pandas as pd
+
+# User-set variables
+BILLING_PROJECT_ID="biodata-catalyst"
```
xbrianh, line 12: Billing project is already available in the env var `GOOGLE_PROJECT`

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
+import io
+import os
+import json # might not be needed
+from firecloud import fiss
+from firecloud.errors import FireCloudServerError
+import firecloud.api as fapi
+import numpy as np
+import pandas as pd
+import logging
+#import re <-- doesn't work on Terra
```
xbrianh, line 10: This worked for me. Make sure your notebook environment is Python 3, not Python 2

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+
+# User-set variables
+BILLING_PROJECT_ID="biodata-catalyst"
```
xbrianh, line 13: If users are executing this code in a notebook, billing project is already available in the env var `GOOGLE_PROJECT`

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+WORKSPACE="TSV-AFY"
+SUBDIRECTORY="\/case2\/"
```
xbrianh, line 15: I don't yet understand the need for escapes. They can probably be avoided.

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+
+# Call FireCloud API
+try:
+    bucket = os.environ["WORKSPACE_BUCKET"]
+    response = fapi.list_entity_types(BILLING_PROJECT_ID, WORKSPACE)
+    if response.status_code != 200:
+        print("Error in Firecloud, check your billing project ID and the name of your workspace.")
+    else:
+        print("Firecloud has found your workspace!")
+        directory = bucket + SUBDIRECTORY
```
xbrianh, line 38: "directory" is a bit of a misnomer, since buckets are not filesystems. I'd generally use "prefix" or "key_prefix" instead.

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+# Make sure to escape the slashes in your SUBDIRECTORY variable. 
+# For instance, if your files are in a folder called "testfiles" you will need to
+# enter it as "/\testfiles/\" or else Python will get angry.
+
+# Call FireCloud API
+try:
+    bucket = os.environ["WORKSPACE_BUCKET"]
+    response = fapi.list_entity_types(BILLING_PROJECT_ID, WORKSPACE)
+    if response.status_code != 200:
+        print("Error in Firecloud, check your billing project ID and the name of your workspace.")
```
xbrianh, line 35: Consider raising an exception, which would halt execution:
```
raise Exception("Error in Firecloud, check your billing project ID and the name of your workspace.")
```

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+    else:
+        print("Firecloud has found your workspace!")
+        directory = bucket + SUBDIRECTORY
+except NameError:
+    print("Caught a NameError exception. This probably means you didn't restart the kernal after"
+          " running the first block of code (the one with all the imports). Run it again, restart"
+          " the kernal, then try running every block of code (including the import one) again.")
```
xbrianh, line 42: You probably want to halt execution here as well. This can be done by changing `print` to `Exception`, or adding a `raise` to trigger the original name error:
```
try:
    stuff()
except NameError:
    print("...")
    raise
```

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+
+# Do magic to create a TSV file
+# We don't have the option to use Python's parsing directly, as the notebook VM is seperate 
+# from the Google Cloud bucket where the data is actually stored, and we cannot run arbitrary 
+# code in a Google Cloud bucket. This forces us to parse the output of gsutil ls. 
+# Parsing ls is always a little dodgy, but we really don't have any other option. 
+# As a result, if your filenames contain non-ascii (ie, stuff besides A-Z, a-z, underscores, and dashes)
+# or bizarre characters (ie, newlines) there is a chance this will not work as expected.
```
xbrianh, line 50: Buckets can be listed using Python:
```
from google.cloud.storage import Client
client = Client()
bucket = client.bucket("my_bucket_name")
for item in bucket.list_blobs(prefix="my_prefix"):
    print(item.name)
```

scripts/file-finder.py
```diff
@@ -0,0 +1,52 @@
...
+        print("Error in Firecloud, check your billing project ID and the name of your workspace.")
+    else:
+        print("Firecloud has found your workspace!")
+        directory = bucket + SUBDIRECTORY
+except NameError:
+    print("Caught a NameError exception. This probably means you didn't restart the kernal after"
+          " running the first block of code (the one with all the imports). Run it again, restart"
+          " the kernal, then try running every block of code (including the import one) again.")
+
+# Append contents.txt with ls
```
DailyDreaming, line 34: The mixed code might be confusing?  Also, generating the extra files can be avoided.  To do this in python rather than bash would be something like:

```
import google.cloud.storage

storage_client = google.cloud.storage.Client()

with open('final.tsv', 'w') as f:
    f.write(f"entity:{TABLE_NAME}_id\tfile_location\n")
    for file_path in return storage_client.list_blobs(bucket_name='wesout', prefix='TopMedAligner'):
        f.write(f'{file_path.split("/")[-1]}\t{file_path}\n')
```

scripts/file-finder.py
```diff
@@ -0,0 +1,52 @@
...
+!gsutil ls $directory > contentlocations.txt
+# Append each line with their file names + full address 
+# of where the files live in your google bucket
+!cat contentlocations.txt | sed 's@.*/@@' > filenames.txt
+!paste filenames.txt contentlocations.txt > combined.txt
+# Set up header that Terra requires for data tables
+# Simply doing this in Python, ie, 
+#headerstring = "entity:" + TABLE_NAME + "_id\tfile_location"
+# results in the tab being converted to a space. Not sure why
+!touch temp.txt
```
DailyDreaming, line 44: `!touch temp.txt` is unnecessary here.  `!echo "entity:$TABLE_NAME""_id\tfile_location" >> temp.txt` will create the file.

scripts/file-finder.py
```diff
@@ -0,0 +1,52 @@
...
+# Make sure to escape the slashes in your SUBDIRECTORY variable. For instance, 
+# if your files are in a folder called "testfiles" you will need to enter it 
+# as "/\testfiles/\" or else Python will get angry.
+
+# Call FireCloud API
+try:
+    bucket = os.environ["WORKSPACE_BUCKET"]
+    response = fapi.list_entity_types(BILLING_PROJECT_ID, WORKSPACE)
+    if response.status_code != 200:
+        print("Error in Firecloud, check your billing project ID and the name of your workspace.")
```
DailyDreaming, line 25: This should raise an Exception here, otherwise this will continue to run code.

scripts/file-finder.py
```diff
@@ -0,0 +1,52 @@
...
+    else:
+        print("Firecloud has found your workspace!")
+        directory = bucket + SUBDIRECTORY
+except NameError:
+    print("Caught a NameError exception. This probably means you didn't restart the kernal after"
+          " running the first block of code (the one with all the imports). Run it again, restart"
+          " the kernal, then try running every block of code (including the import one) again.")
```
DailyDreaming, line 32: Same, raise an exception here rather than swallow the error, otherwise the code will continue to run.

Might be a good place for a custom error too.  Would look something like:

```
class KernelReloadException(Exception):
    def __init__(self):
        msg = "Please try restarting the kernel."
        super().__init__(msg)

raise KernelReloadException
```

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
+import io
```
DailyDreaming, line 1: Again, the imports are a little heavy here.  You really only use:
```
import os
import firecloud.api as fapi
import pandas as pd
import logging
```

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+# Parsing ls is always a little dodgy, but we really don't have any other option. 
+# As a result, if your filenames contain non-ascii (ie, stuff besides A-Z, a-z, underscores, and dashes)
+# or bizarre characters (ie, newlines) there is a chance this will not work as expected.
+
+# Pandas dataframe method
+# Note: I am sure there is a more efficient way of doing this!
+
+def baseID(filename_string, child_extension):
+    global PARENT_FILETYPE
+    global INCLUDE_PARENT_EXTENSION
```
DailyDreaming, line 57: These globals are not necessary to declare.  They're already declared above the function so they're within the scope of what the function can see.

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+
+# Get location of everything and their file names
+logging.info("Querying Google and writing results...")
+!gsutil ls $directory > contentlocations.txt
+!cat contentlocations.txt | sed 's@.*/@@' > filenames.txt
+
+# Import everything
+logging.info("Constructing dataframe...")
+data={}
+this_file=open("contentlocations.txt", "r")
```
DailyDreaming, line 72: Should be using a context manager here, specifically:
`with open(filename, 'r') as f`

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+    file1.close()
+full_header="entity:"+TABLE_NAME+"_id"+header
+with open('final.tsv', "a") as file2:
+    file2.write(full_header)
+    for string in everything_else:
+        file2.write(string)
+    file2.close()
+
+# Clean up
+!rm dataframe.tsv
```
DailyDreaming, line 169: You can use `os.remove("dataframe.tsv")`

scripts/paternity-test.py
aofarrel: Without escapes it will cause errors later down the line.

```
BILLING_PROJECT_ID="biodata-catalyst"
WORKSPACE="TSV-AFY -- Case 4 -- Top Cekrit"
SUBDIRECTORY="/case1/"
TABLE_NAME="CRAMs"
try:
    bucket = os.environ["WORKSPACE_BUCKET"]
    response = fapi.list_entity_types(BILLING_PROJECT_ID, WORKSPACE)
    if response.status_code != 200:
        print("Error in Firecloud, check your billing project ID and the name of your workspace.")
    else:
        print("Firecloud has found your workspace!")
        directory = bucket + SUBDIRECTORY
except NameError:
    print("Caught a NameError exception. This probably means you didn't restart the kernal after"
          " running the first block of code (the one with all the imports). Run it again, restart"
          " the kernal, then try running every block of code (including the import one) again.")
!gsutil ls $directory
```
Firecloud will report having found the workspace as it doesn't look for the subdirectory, but gstutil ls will throw a command error. \case1\ on the other hand will not even run as thinks we're trying to escape the end of the string.

scripts/file-finder.py
```diff
@@ -0,0 +1,52 @@
+import io
```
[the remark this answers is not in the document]
aofarrel: dammit, I knew I left in something I wasn't using! thanks, not sure how I forgot to check that.

scripts/paternity-test.py
```diff
@@ -0,0 +1,173 @@
...
+for child_extension in unique_children: #loop once per child extension
+    i=i+1
+    logging.info("\t* Dealing with %s (child %d out of %d)" % (child_extension, i, len(unique_children)))
+    df[child_extension] = ""
+    df[child_extension+"_location"] = ""
+    output_list={}
+    for index_label, row_series in df.iterrows():
+        # On first iteration, iterate over parent files
+        # Only do once to avoid wasting time repeating this constantly
+        if (df.at[index_label,'FileType'] == PARENT_FILETYPE) & (i==1):
```
DailyDreaming, line 105: Are sure you want a bitwise `&` here?

```
>>> 1 & 2
0
>>> 1 & 0
0
>>> 1 & 1
1
>>> 1 & 111
1
```

scripts/paternity-test.py
aofarrel: For future reference, is there any advantage to this or is this just to keep things pythonic (ie avoid bash)?

DailyDreaming: Just to keep things pythonic.

xbrianh:
@aofarrel Closing this pull request. We can go through another round of reviews with a new PR if needed.

PR #18, opened 2020-04-08
Add head_vcf.py script

xbrianh:


scripts/head_vcf.py
```diff
@@ -0,0 +1,29 @@
...
+    raw = gscio.Reader(blob)
+else:
+    raw = open(args.cloudpath, "rb")
+
+with closing(raw):
+    with bgzip.BGZipReader(raw) as bgreader:
+        with io.BufferedReader(bgreader) as reader:
+            for line in reader:
+                if line.startswith(b"#"):
+                    print(line.decode("ascii").strip())
```
DailyDreaming, line 27: Are these guaranteed to be `ascii`?

scripts/head_vcf.py
```diff
@@ -0,0 +1,29 @@
...
+    blob = Client().bucket(bucket_name).get_blob(key)
+    raw = gscio.Reader(blob)
+else:
+    raw = open(args.cloudpath, "rb")
+
+with closing(raw):
+    with bgzip.BGZipReader(raw) as bgreader:
+        with io.BufferedReader(bgreader) as reader:
+            for line in reader:
+                if line.startswith(b"#"):
```
DailyDreaming, line 26: I could be misremembering, but I feel like I've seen something like:

```
# header header
# more header

# casual line skipped
actual vcf contents
```

Would it be good to also skip `line.strip() == ''` and only break if it hits content?

scripts/head_vcf.py
xbrianh: Actually, this should be utf-8 according to [the spec](http://samtools.github.io/hts-specs/VCFv4.3.pdf)

Thanks!

scripts/head_vcf.py
xbrianh: will do

xbrianh:
@DailyDreaming I'm going to merge this based on manual testing, but see [this issue](https://github.com/DataBiosphere/terra-notebook-utils/issues/19)

DailyDreaming:
@xbrianh Sounds good!

PR #20, opened 2020-04-09
Add functions useful for managing Terra data tables

xbrianh:


terra_notebook_utils/table.py
```diff
@@ -39,6 +39,26 @@ def fetch_drs_url(table: str, file_name: str):
         raise ValueError(f"Expected DRS url in {table} for {file_name}, got {val} instead.")
     return val
 
+def list_entities(ent_type):
+    resp = fiss.fapi.get_entities_with_type(WORKSPACE_GOOGLE_PROJECT, WORKSPACE_NAME)
```
DailyDreaming, line 43: Think we'll ever need the paginated version? https://github.com/broadinstitute/fiss/blob/46742e1ab7cbdc1b18e962d009a4a59715ade894/firecloud/api.py#L450

terra_notebook_utils/table.py
```diff
@@ -39,6 +39,26 @@ def fetch_drs_url(table: str, file_name: str):
...
+            yield ent
+
+def delete_entities(entities: list):
+    ents = [dict(entityType=e['entityType'], entityName=e['name'])
+            for e in entities]
+    resp = fiss.fapi.delete_entities(WORKSPACE_GOOGLE_PROJECT, WORKSPACE_NAME, ents)
+    resp.raise_for_status()
+
+def delete_table(ent_type: str):
+    delete_entities([e for e in list_entities(ent_type)])
```
DailyDreaming, line 56: Will we ever need to batch this?

terra_notebook_utils/table.py
xbrianh: Possibly, but maybe not right now.

I'll add an issue.

xbrianh: See https://github.com/DataBiosphere/terra-notebook-utils/issues/21

terra_notebook_utils/table.py
xbrianh: Future work, see: https://github.com/DataBiosphere/terra-notebook-utils/issues/21

PR #23, opened 2020-04-10
Try harder to find google project env var

xbrianh:
This is useful when running outside of the Terra notebook environment.


xbrianh:
@DailyDreaming This is a small change so I merged immediately, but you can take a post-hoc review if you wish.

PR #27, opened 2020-04-14
Use VCFInfo class to output VCF header

xbrianh:
This makes it easier to view VCF headers from a notebook.


xbrianh:
@DailyDreaming Yes, we should bring in typing. Maybe one of the new interns :)

PR #28, opened 2020-04-14
[Easy] Use "git clean" during "make clean"

xbrianh:


xbrianh:
@DailyDreaming This is simple so I'm merging immediately. Post-hoc reviews are welcome.

PR #30, opened 2020-04-14
Add test mode, introduce gitlab CI

xbrianh:


tests/infra/testmode.py
```diff
@@ -0,0 +1,7 @@
+import os
+import unittest
+
+
+def controlled_access(f):
```
DailyDreaming, line 5: Seems like a good idea and I'll likely add something like this as well, until I'm guessing our code bases share enough that it makes sense to merge.

PR #31, opened 2020-04-14
Include samples in VCFInfo header

xbrianh:
Add "samples" property to VCFInfo


xbrianh:
Closing this in favor of pure python implementation for now.

PR #36, opened 2020-04-27
Bundle htslib and bcftools

xbrianh:
The bundled samtools software includes [htslib](https://github.com/samtools/htslib) and [bcftools](https://github.com/samtools/bcftools), which have been modified to merge multisample, completely aligned VCFs without requiring index files.


terra_notebook_utils/samtools.py
```diff
@@ -0,0 +1,25 @@
...
+
+bin_root = os.path.abspath(os.path.dirname(sys.executable))
+htsfile_path = os.path.join(bin_root, "htsfile")
+bcftools_path = os.path.join(bin_root, "bcftools")
+bcftools_available = os.path.isfile(bcftools_path)
+
+try:
+    p = _run([htsfile_path, "--version"], capture_output=True)
+    htsfile_available = True
+except (FileNotFoundError, subprocess.CalledProcessError):
```
DailyDreaming, line 18: Maybe something like:

```
except (FileNotFoundError, subprocess.CalledProcessError) as e:
    print(e)  # or traceback.print_exc()
    htsfile_available = False
```

xbrianh: An exception here indicates htlib failed to build during `setup.py`. Printing out the trace would not help diagnose that problem, and would therefore be confusing.

We could do something like:
```
except (FileNotFoundError, subprocess.CalledProcessError) as e:
    print("htsfile unavailable: htslib build failed during installation. try `pip install -v terra-notebook-utils` to diagnose the problem")
    htsfile_available = False
```

xbrianh: done.

xbrianh:
Closed in favor of a separate pip installable package [xsamtools](https://github.com/xbrianh/xsamtools)

PR #57, opened 2020-06-01
Simplify cli configuration

xbrianh:


terra_notebook_utils/cli/config.py
```diff
@@ -10,15 +10,24 @@
 config_cli = dispatch.group("config", help=__doc__)
 
 
-@config_cli.command("set", arguments={
-    "--key": dict(choices=Config.info.keys()),
-    "--value": dict()
+@config_cli.command("set-workspace", arguments={
+    "value": dict()
```
DailyDreaming, line 14: `value` might be a bit ambiguous?

terra_notebook_utils/cli/config.py
```diff
@@ -10,15 +10,24 @@
...
     """
-    Set a configuration value. Use `tnu config print` to see current settings.
+    Set workspace for cli commands
     """
-    Config.info[args.key] = args.value
+    Config.info["workspace"] = args.value
+    Config.write()
+
+@config_cli.command("set-workspace-google-project", arguments={
+    "value": dict()
```
DailyDreaming, line 24: Same.  Maybe `project`?

terra_notebook_utils/cli/config.py
xbrianh: good catch, will fix both

PR #59, opened 2020-06-01
Suppress GCP and socket warnings during tests

xbrianh:
This makes it easier to spot useful test output.


xbrianh:
@DailyDreaming Just FYI: the unittest framework resets warnings for each test, so you need to put this in `def setUp` to work correctly.

PR #60, opened 2020-06-03
Update DRS CLI and API

xbrianh:


xbrianh:
@DailyDreaming Merging now to make it into a release for the morning, but a post-hoc review is welcome.

PR #69, opened 2020-06-15
[IA-1919] Adding requester pays and ENV support. Documentation added as well

jdcanas:
This adds support for allowing drs resolution for buckets that have requester pays enabled. 

For some context, copying from requester pays buckets requires the service account that bond has to be enabled through the rawls url that is called. It is implemented such that requester pays is lazily enabled before the first martha call occurs. If it is already enabled, this is a no-op. If it fails, the martha call is still executed, but prior to that a message is printed to the user letting them know they won't be able to copy from URIs that require requester.


terra_notebook_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.0'
+__version__ = '0.1.0'
```
jdcanas, line 1: is this actually used? When I pip install this, it tells me its on version `0.1.1`

xbrianh: No, this is unused (good catch!). See https://github.com/DataBiosphere/terra-notebook-utils/pull/69#discussion_r440376144

terra_notebook_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.0'
+__version__ = '0.2.1'
```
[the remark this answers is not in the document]
jdcanas: I took a stab at it

terra_notebook_utils/drs.py
```diff
@@ -17,48 +18,77 @@ def _parse_gs_url(gs_url):
...
+
     if 200 == resp.status_code:
         resp_data = resp.json()
     else:
         print(resp.content)
         raise Exception(f"expected status 200, got {resp.status_code}")
     return resp_data
 
-def resolve_drs_for_gs_storage(drs_url):
+def resolve_drs_for_gs_storage(drs_url, env: str=ENV):
```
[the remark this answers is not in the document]
jdcanas: good catch. I originally had it here for testing requester pays logic without being locked into 1 env for all tests, but it proved too complicated because it requires a specific credentials to test out. I made a follow-up ticket to test it out closer to the systems that support the functionality.

CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+## 0.0.1 - 06/12/2020
+- Added requester pays set-up to drs module
+- Added ENV variable for local testing to __init__
+- Added some print messages to drs module to let user know what is happening
+
+## 0.0.0 - 01/29/2020
+- initial release completed
```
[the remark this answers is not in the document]
jdcanas: Not too sure how to do this, but I took a stab at resolving the version.py updating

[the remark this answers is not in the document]
xbrianh: The machinery is actually already available in [common.mk](https://github.com/DataBiosphere/terra-notebook-utils/blob/7b52c130d5065c3df4d284a74173b7a310ceda34/common.mk#L28). All that is needed is s/Changes.md/CHANGELOG.md

terra_notebook_utils/drs.py
```diff
@@ -17,21 +20,45 @@ def _parse_gs_url(gs_url):
     else:
         raise RuntimeError(f'Invalid gs url schema.  {gs_url} does not start with {_GS_SCHEMA}')
 
+@functools.lru_cache(maxsize=1)
```
xbrianh, line 23: Now that there are no parameters ,`maxsize=1` can be removed.

terra_notebook_utils/drs.py
```diff
@@ -4,11 +4,14 @@
 import json
 import requests
 
-from terra_notebook_utils import WORKSPACE_GOOGLE_PROJECT, WORKSPACE_BUCKET, _GS_SCHEMA
+from terra_notebook_utils import WORKSPACE_GOOGLE_PROJECT, WORKSPACE_BUCKET, TERRA_DEPLOYMENT_ENV, WORKSPACE_NAME, _GS_SCHEMA # noqa
```
xbrianh, line 7: This can be split into two lines to avoid the `# noqa`:
```
from terra_notebook_utils import (WORKSPACE_GOOGLE_PROJECT, WORKSPACE_BUCKET, TERRA_DEPLOYMENT_ENV, WORKSPACE_NAME,
                                  _GS_SCHEMA)
```

terra_notebook_utils/drs.py
```diff
@@ -17,21 +20,45 @@ def _parse_gs_url(gs_url):
     else:
         raise RuntimeError(f'Invalid gs url schema.  {gs_url} does not start with {_GS_SCHEMA}')
 
+@functools.lru_cache(maxsize=1)
+def enable_requester_pays():
+
```
xbrianh, line 25: Please remove this blank line.

PR #71, opened 2020-06-15
Fix versioning

### LINKED ISSUE #70: Fix release versioning
Version should be tracked in `terra_notebook_utils/version.py`, and updated automatically during `make release_*`. `setup.py` should read from `terra_notebook_utils/version.py`.

See discussion [here](https://packaging.python.org/guides/single-sourcing-package-version/).

xbrianh:
closes #70


terra_notebook_utils/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.0'
+__version__ = '0.1.1'
```
DailyDreaming, line 1: Maybe add a comment here to not edit/bump this directly but to always run `make release`?

xbrianh: Good idea! Will add.

PR #73, opened 2020-06-18
Introduce method to resolve CLI config

xbrianh:
This allows the CLI to fallback to environment variable configurations. The default heierarchy is:
  - First use values passed in to a CLI command
  - Otherwise use values from the CLI config file
  - Otherwise use environment variable values

depends on #72

### DIFF
--- a/terra_notebook_utils/cli/__init__.py
+++ b/terra_notebook_utils/cli/__init__.py
@@ -7,6 +7,8 @@ import typing
 
 import cli_builder
 
+from terra_notebook_utils import WORKSPACE_NAME, WORKSPACE_GOOGLE_PROJECT
+
 
 class Config:
     info = dict(workspace=None, workspace_google_project=None)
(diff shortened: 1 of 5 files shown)


terra_notebook_utils/cli/__init__.py
```diff
@@ -23,6 +25,15 @@ def load(cls):
     def write(cls):
         with open(cls.path, "w") as fh:
             fh.write(json.dumps(cls.info, indent=2))
+
+    @classmethod
+    def resolve(cls, default_workspace: str, default_namespace: str):
+        workspace = default_workspace or (cls.info['workspace'] or WORKSPACE_NAME)
+        namespace = default_namespace or (cls.info['workspace_google_project'] or WORKSPACE_GOOGLE_PROJECT)
```
DailyDreaming, line 32: I'd expect a default to be the last item resolved in a short circuit.  This looks like it will always choose the default if it exists, is that correct?

```
>>> i = '' or None or 'Best Choice' or 'default'
>>> print(i)
Best Choice
```

terra_notebook_utils/cli/__init__.py
```diff
@@ -23,6 +25,15 @@ def load(cls):
...
+        if workspace and namespace is None:
+            from terra_notebook_utils.workspace import get_workspace_namespace
```
DailyDreaming, line 34: Why is this imported inside of the method?  Just a nit I suppose.

terra_notebook_utils/cli/__init__.py
xbrianh: I'll change the names away from "default_*"

terra_notebook_utils/cli/__init__.py
xbrianh: Loading modules lazily may improve the responsiveness of the CLI.

PR #82, opened 2020-06-30
Propagate workspace/namespace args to drs methods

xbrianh:
This allows requester pays functionality to work with CLI commands.


DailyDreaming:
@xbrianh `ModuleNotFoundError: No module named 'oauth2client'`

PR #100, opened 2020-07-31
Introduce DRS batch copy API

xbrianh:
depends on #99

### DIFF
--- a/terra_notebook_utils/drs.py
+++ b/terra_notebook_utils/drs.py
@@ -6,14 +6,16 @@ import re
 import json
 import logging
 import requests
+from concurrent.futures import ThreadPoolExecutor
 from functools import lru_cache
 from collections import namedtuple
-from typing import Tuple, Optional
+from typing import Tuple, Iterable, Optional
 
-from terra_notebook_utils import WORKSPACE_GOOGLE_PROJECT, WORKSPACE_BUCKET, WORKSPACE_NAME
+from terra_notebook_utils import WORKSPACE_GOOGLE_PROJECT, WORKSPACE_BUCKET, WORKSPACE_NAME, MULTIPART_THRESHOLD
 from terra_notebook_utils import gs, tar_gz, TERRA_DEPLOYMENT_ENV, _GS_SCHEMA
 
 import gs_chunked_io as gscio
+from gs_chunked_io import async_collections
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


terra_notebook_utils/drs.py
```diff
@@ -156,6 +158,32 @@ def copy(drs_url: str,
     else:
         copy_to_local(drs_url, dst, workspace_name, google_billing_project)
 
+def copy_batch(drs_urls: Iterable[str],
+               dst: str,
+               workspace_name: Optional[str]=WORKSPACE_NAME,
+               google_billing_project: Optional[str]=WORKSPACE_GOOGLE_PROJECT):
+    enable_requester_pays(workspace_name, google_billing_project)
+    with ThreadPoolExecutor(max_workers=3) as oneshot_executor:
```
DailyDreaming, line 166: Max workers looks like a magic number here.  Should `ThreadPoolExecutor(max_workers=3)` and `AsyncSet(oneshot_executor, concurrency=3)` be set with a default variable (as with MULTIPART_THRESHOLD?).  Or an input arg with a default of 3?  And/or a function of the number of cores the machine has?

tests/test_terra_notebook_utils.py
```diff
@@ -116,6 +118,30 @@ def test_copy(self):
             key = f"gs://{WORKSPACE_BUCKET}/test_oneshot_object_{uuid4()}"
             drs.copy(self.drs_url, key)
 
+    @testmode("controlled_access")
+    def test_copy_batch(self):
+        drs_urls = {
+            "NWD522743.b38.irc.v1.cram.crai": "drs://dg.4503/95cc4ae1-dee7-4266-8b97-77cf46d83d35",  # 1631686 bytes
+            # "NWD961306.freeze5.v1.vcf.gz": "drs://dg.4503/6e73a376-f7fd-47ed-ac99-0567bb5a5993",  # 2679331445 bytes
```
DailyDreaming, line 125: Temporarily inaccessible?  XD  The size is a good comment.

terra_notebook_utils/drs.py
xbrianh: good point. will fix.

tests/test_terra_notebook_utils.py
xbrianh: Including the 2GB files increases the duration of the test, but they're still useful for spot testing large files now and then.

PR #102, opened 2020-08-05
Fix bump gs-chunked-io error in tarball extraction

xbrianh:


tests/test_terra_notebook_utils.py
```diff
@@ -144,7 +144,14 @@ def test_copy_batch(self):
 
     # Probably don't want to run this test very often. Once a week?
     def _test_extract_tar_gz(self):
-        drs_url = "drs://dg.4503/273f3453-4d16-4ddd-8877-dbac958a4f4d"  # Amish cohort v4 VCF
+        drs_url = "drs://dg.4503/273f3453-4d16-4ddd-8877-dbac958a4f4d"  # Amish cohort v4 VCF  # no access
```
DailyDreaming, line 147: Should all of these be comments?  The one good one, should it be a part of a tested list?

PR #106, opened 2020-08-31
Add a tnu drs head bytes function.

DailyDreaming:
See: https://ucsc-cgl.atlassian.net/browse/TLP-1535

@xbrianh 

I was worried if I just called this `head`, then people might expect to be able to read the first 10 characters/lines of a DRS object, so I renamed it `check_accessible` to better fit the use-case.  I could make this into an actual `head` function, but I'm worried that that would entail not using `raw_read`, which means decompressing on the Google side, so it would likely slow things down if we're only checking for accessibility.

Thoughts?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-google-cloud-storage>=1.31.0
+google-cloud-storage==1.31.0
 gs-chunked-io >= 0.4, < 0.5
 firecloud
 bgzip >= 0.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-google-cloud-storage
+google-cloud-storage>=1.31.0
 gs-chunked-io >= 0.4, < 0.5
 firecloud
 bgzip >= 0.3.0
--- a/terra_notebook_utils/cli/drs.py
+++ b/terra_notebook_utils/cli/drs.py
@@ -56,20 +56,27 @@ def drs_copy_batch(args: argparse.Namespace):
     drs.copy_batch(args.drs_urls, args.dst, args.workspace, args.google_billing_project)
 
 @drs_cli.command("head", arguments={
-    "-n": dict(type=int, required=False, default=1, help="Return the first integer n bytes of a file (uncompressed)."),
+    "-c": dict(type=int, required=False, default=None,
+               help="Return the first integer n bytes of a file (uncompressed)."),
     "drs_url": dict(type=str),
     ** workspace_args,
 })
 def drs_head(args: argparse.Namespace):
     """
-    Fetch the first byte of a drs:// object to verify access.
+    Print the first bytes of a drs:// object.
 
     Example:
-        tnu drs check-accessible drs://crouching-drs-hidden-access
+        tnu drs head drs://crouching-drs-hidden-access
     """
+    # if the user specified nothing, only fetch the first byte
+    args.c = 1 if args.c is None else args.c
+    assert args.c > 0
+
     args.workspace, args.google_billing_project = Config.resolve(args.workspace, args.google_billing_project)
-    result = drs.head(args.drs_url, args.n, args.workspace, args.google_billing_project)
-    print(result)
+    drs.head(args.drs_url,
+             num_bytes=args.c,
+             workspace_name=args.workspace,
+             google_billing_project=args.google_billing_project)
 
 @drs_cli.command("extract-tar-gz", arguments={
     "drs_url": dict(type=str),
--- a/terra_notebook_utils/cli/drs.py
+++ b/terra_notebook_utils/cli/drs.py
@@ -55,18 +55,20 @@ def drs_copy_batch(args: argparse.Namespace):
     args.workspace, args.google_billing_project = Config.resolve(args.workspace, args.google_billing_project)
     drs.copy_batch(args.drs_urls, args.dst, args.workspace, args.google_billing_project)
 
-@drs_cli.command("check-accessible", arguments={
+@drs_cli.command("head", arguments={
+    "-n": dict(type=int, required=False, default=1, help="Return the first integer n bytes of a file (uncompressed)."),
     "drs_url": dict(type=str),
     ** workspace_args,
 })
-def drs_check_accessible(args: argparse.Namespace):
+def drs_head(args: argparse.Namespace):
     """
-    Fetch the first byte of a drs:// object to verify access
-    example:
+    Fetch the first byte of a drs:// object to verify access.
+
+    Example:
         tnu drs check-accessible drs://crouching-drs-hidden-access
     """
     args.workspace, args.google_billing_project = Config.resolve(args.workspace, args.google_billing_project)
-    result = drs.check_accessible(args.drs_url)
+    result = drs.head(args.drs_url, args.n, args.workspace, args.google_billing_project)
     print(result)
 
 @drs_cli.command("extract-tar-gz", arguments={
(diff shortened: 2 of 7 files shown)


xbrianh:
One possible name for a function that outputs a single byte is `peak`.

But perhaps it's fine to call it "head" and output the first n byte. Is there any BDCat data that uses the gzip compression features of Google storage? We can document the expectation that, most of the time, "head" output would be binary.

For some savvy users it may be that heading out the first MB of a binary file is exactly what they want to do.

terra_notebook_utils/drs.py
```diff
@@ -121,6 +125,25 @@ def copy_to_local(drs_url: str,
...
+    """
+    Check access to a DRS object by attempting to access its first byte of data.
+    """
+    assert drs_url.startswith("drs://")
+    enable_requester_pays(workspace_name, google_billing_project)
+    client, info = resolve_drs_for_gs_storage(drs_url)
+    blob = client.bucket(info.bucket_name, user_project=google_billing_project).blob(info.key)
+    try:
+        # don't expand compressed files, to save time
+        blob.download_as_string(start=0, end=1, raw_download=True)
```
xbrianh, line 140: `download_as_string` is depreciated.

Try `download_as_bytes`

terra_notebook_utils/cli/drs.py
```diff
@@ -55,6 +55,20 @@ def drs_copy_batch(args: argparse.Namespace):
...
+    ** workspace_args,
+})
+def drs_check_accessible(args: argparse.Namespace):
+    """
+    Fetch the first byte of a drs:// object to verify access
+    example:
+        tnu drs check-accessible drs://crouching-drs-hidden-access
+    """
+    args.workspace, args.google_billing_project = Config.resolve(args.workspace, args.google_billing_project)
+    result = drs.check_accessible(args.drs_url)
```
xbrianh, line 69: Both `args.workspace`, and `args.google_billing_project` need to be passed into `check_accessible`, otherwise this will fail for requester pays buckets.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-google-cloud-storage
+google-cloud-storage>=1.31.0
```
xbrianh, line 1: Can we add an upper pin to protect TNU from future breaking changes to google-cloud-storage?

terra_notebook_utils/cli/drs.py
```diff
@@ -55,6 +55,22 @@ def drs_copy_batch(args: argparse.Namespace):
     args.workspace, args.google_billing_project = Config.resolve(args.workspace, args.google_billing_project)
     drs.copy_batch(args.drs_urls, args.dst, args.workspace, args.google_billing_project)
 
+@drs_cli.command("head", arguments={
+    "-n": dict(type=int, required=False, default=1, help="Return the first integer n bytes of a file (uncompressed)."),
```
xbrianh, line 59: Considering matching CLI argument documentation for unix `head`:
```
  -c, --bytes=[-]NUM       print the first NUM bytes of each file;
                             with the leading '-', print all but the last
                             NUM bytes of each file
  -n, --lines=[-]NUM       print the first NUM lines instead of the first 10;
                             with the leading '-', print all but the last
                             NUM lines of each file
```

terra_notebook_utils/drs.py
```diff
@@ -121,6 +125,25 @@ def copy_to_local(drs_url: str,
...
+    Check access to a DRS object by attempting to access its first byte of data.
+    """
+    assert drs_url.startswith("drs://")
+    enable_requester_pays(workspace_name, google_billing_project)
+    client, info = resolve_drs_for_gs_storage(drs_url)
+    blob = client.bucket(info.bucket_name, user_project=google_billing_project).blob(info.key)
+    try:
+        # don't expand compressed files, to save time
+        return blob.download_as_bytes(start=0, end=n, raw_download=True)
+    except Exception:
```
xbrianh, line 142: This clause should catch exceptions related to an inaccessible object. Other exceptions should be raised normally.

Please consider:
```
except (google.cloud.exceptions.NotFound, google.cloud.exceptions.Forbidden):
```

terra_notebook_utils/drs.py
```diff
@@ -121,6 +125,25 @@ def copy_to_local(drs_url: str,
...
+    """
+    Check access to a DRS object by attempting to access its first byte of data.
+    """
+    assert drs_url.startswith("drs://")
+    enable_requester_pays(workspace_name, google_billing_project)
+    client, info = resolve_drs_for_gs_storage(drs_url)
+    blob = client.bucket(info.bucket_name, user_project=google_billing_project).blob(info.key)
+    try:
+        # don't expand compressed files, to save time
+        return blob.download_as_bytes(start=0, end=n, raw_download=True)
```
xbrianh, line 141: Consider dropping `raw_download`:
  - As far as I'm aware, there are no BDCat files that meet the requirements for decompressive transcoding
  - If there are such files, returning decompressed bytes will be more in line with user expectations
  - Decompressing the first few bytes of a file is not a large time penalty

See documentation for Google decompressive transcoding:
https://cloud.google.com/storage/docs/transcoding

terra_notebook_utils/drs.py
```diff
@@ -23,6 +24,9 @@
 
 DRSInfo = namedtuple("DRSInfo", "credentials bucket_name key name size updated")
 
+class InaccessibleDrsUrlException(Exception):
```
xbrianh, line 27: Consider renaming to something like `GSBlobInnaccessible`. In a future PR, we might consider raising this error for all drs:// and gs:// URLs that fail to resolve because of non-existence or acces-denied.

terra_notebook_utils/cli/drs.py
DailyDreaming: See also https://github.com/DataBiosphere/terra-notebook-utils/pull/107

requirements.txt
DailyDreaming: I just pinned 1.31.0, since it's the latest.

terra_notebook_utils/drs.py
```diff
@@ -121,6 +127,55 @@ def copy_to_local(drs_url: str,
...
+    :param blob: The Google blob object.
+    :param num_bytes: Number of bytes to print before stopping.
+    :param buffer: Size of the data chunks to grab each time.  The default buffer is 32Mib.
+    """
+    num_bytes_left = num_bytes
+    start = 0
+    end = buffer
+    while True:
+        if num_bytes_left <= buffer:
+            print(str(blob.download_as_bytes(start=start, end=num_bytes), sys.stdout.encoding), end='')
```
xbrianh, line 145: Does converting to `str` transform the contents?

For instance, I'd be surprised if `headed_binary` contained a decoded string instead of the first 1024 bytes of my DRS URI:
```
tnu drs head -c 1024 drs://crouching-drs-hidden-access > headed_binary
```

DailyDreaming: Hmmm... I should write a test for it, but it should have the same behavior as the actual head function since I use `sys.stdout.encoding` which tells str() how to process it to stdout.

The purpose was to remove the extraneous `b""` wrapper, which I don't feel is good to output, and adds extra characters if piped to a file.

I'll see if there's a better way and what the actual head function does.

DailyDreaming:
@xbrianh I think that this should be okay for re-review.

terra_notebook_utils/drs.py
```diff
@@ -121,6 +127,56 @@ def copy_to_local(drs_url: str,
         logger.info(f"Downloading {drs_url} to {filepath}")
         blob.download_to_file(fh)
 
+def print_bytes(source: Callable, num_bytes: int, buffer: int = MULTIPART_THRESHOLD):
```
xbrianh, line 130: Consider passing in typing.IO instead of typing.Callable. This will allow you to use standard io `read`, and avoids the chunking arithmetic. It also avoids the implicit requirements on `source`, which must support the `start` and `end` parameters.

You can pass in a readable stream over a GS blob with:
```
with gscio.Reader(blob) as handle:
    print_bytes(handle, num_bytes, buffer)
```

xbrianh (approved):
Approving, but please see comment.

PR #108, opened 2020-09-14
Make API load config as well.

DailyDreaming:
Only the CLI was loading the config.  This makes it so that both the CLI and the API load the `~/.tnu_config` file.

I've also incorporated some of the exterior settings (fetching from env vars) that were done outside of config, inside of the config.

### DIFF
--- a/terra_notebook_utils/__init__.py
+++ b/terra_notebook_utils/__init__.py
@@ -23,11 +23,11 @@ class Config:
     @classmethod
     def resolve(cls, override_workspace: Optional[str] = None, override_namespace: Optional[str] = None):
         workspace_name = (override_workspace or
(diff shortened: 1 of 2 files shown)


xbrianh:
This package is used in three general environments: Jupyter notebooks, from the CLI, and Python scripts. I am concerned that loading the CLI config automatically may cause conflicts with the Jupyter notebook environment presented by Terra. I am also concerned that it would make it easier for identical scripts to behave differently for different users. It would be preferable for scripts to either explicitly pick up, or hard code, the Terra workspace and billing project.

If this functionality is desirable, we could consider a convenience method to load the CLI config `load_cli_config()`.

DailyDreaming:
@xbrianh Sold.

PR #109, opened 2020-09-14
Make tnu drs head compatible with python notebooks.

DailyDreaming:
This should make it work.  >.>

In the notebook, stdout and stderr are ipykernel.iostream.OutStream objects:
https://github.com/ipython/ipykernel/blob/master/ipykernel/iostream.py#L265
and seem to be able to handle bytes as well as unicode without the buffer attribute.

### DIFF
--- a/terra_notebook_utils/drs.py
+++ b/terra_notebook_utils/drs.py
@@ -3,6 +3,7 @@ Utilities for working with DRS objects
 """
 import os
 import sys
+import io
 import re
 import json
 import logging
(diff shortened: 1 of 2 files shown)


terra_notebook_utils/drs.py
```diff
@@ -153,8 +153,13 @@ def head(drs_url: str,
                                      f'{traceback.format_exc()}')
     blob = client.bucket(info.bucket_name, user_project=google_billing_project).blob(info.key)
     try:
+        # sys.stdout.buffer is used outside of a python notebook
+        # INSIDE of a python notebook, this is a ipykernel.iostream.OutStream object:
+        # https://github.com/ipython/ipykernel/blob/master/ipykernel/iostream.py#L265
+        # this seems to be able to handle bytes as well as unicode
+        stdout_buffer = getattr(sys.stdout, 'buffer', sys.stdout)
```
xbrianh, line 160: Is this object guaranteed to have a `write` method?

xbrianh (approved):
Approving, but please address comment.

terra_notebook_utils/drs.py
DailyDreaming: Yes.

PR #169, opened 2020-10-02
Remove prepare_merge_workflow_input

xbrianh:
This input style is incomptable with array input for xvcfmerge


DailyDreaming (approved):
LGTM.

PR #170, opened 2020-10-06
Introduce costs module

xbrianh:


terra_notebook_utils/costs.py
```diff
@@ -0,0 +1,25 @@
+
+MINUTE = 1 / 60  # one minute in units of hours
+
+class GCPCustomN1Cost:
+    # Pricing information: https://cloud.google.com/compute/vm-instance-pricing#n1_custommachinetypepricing
+    on_demand_per_hour_per_cpu = 0.033174
```
DailyDreaming, line 6: Is there a way to fetch this information with an API?

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/costs.py
xbrianh: Google has a billing [catalog API](https://cloud.google.com/blog/products/gcp/introducing-cloud-billing-catalog-api-gcp-pricing-in-real-time). Unfortunately it requires billing SKUs to do cost lookups, and this information is not available in Terra workflow metadata.

PR #171, opened 2020-10-06
Introduce workflow module

xbrianh:


terra_notebook_utils/workflows.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def list_submissions(workspace_name: Optional[str]=WORKSPACE_NAME,
+                     google_billing_project: Optional[str]=WORKSPACE_GOOGLE_PROJECT) -> Generator[dict, None, None]:
+    """
+    List all submissions in the workspace `workspace_name`.
+    """
+    resp = fiss.fapi.list_submissions(google_billing_project, workspace_name)
+    resp.raise_for_status()
+    for s in resp.json():
```
DailyDreaming, line 20: Will their endpoint give a paged response if the number is high enough?

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/workflows.py
xbrianh: There is no indicate in the fiss Python API, or the firecloud API, that the response will be paged:
https://github.com/broadinstitute/fiss/blob/a3a51b7cfd72d7527810b8569d36ba923b7d0b72/firecloud/api.py#L1031-L1042

PR #174, opened 2020-10-06
Refactor CLI test mixin

xbrianh:
This will be useful for a reorg of tests into separate modules.

depends on #172 #173

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,52 @@
+import io
+import os
+import sys
+import argparse
+import subprocess
+from contextlib import redirect_stdout
+from tempfile import NamedTemporaryFile
+from typing import List, Callable
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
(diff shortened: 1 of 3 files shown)


tests/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+import argparse
+import subprocess
+from contextlib import redirect_stdout
+from tempfile import NamedTemporaryFile
+from typing import List, Callable
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from tests import config
```
DailyDreaming, line 13: Unused import?

DailyDreaming (approved):
LGTM.  See comment.

tests/__init__.py
xbrianh: Nope, it configures the test environment.

xbrianh: I'll add a comment.

PR #195, opened 2020-11-05
Improve vcf info tests

xbrianh:
depends on #194

### DIFF
--- a/terra_notebook_utils/vcf.py
+++ b/terra_notebook_utils/vcf.py
@@ -93,10 +93,12 @@ def vcf_info(uri: str,
         client, drs_info = drs.resolve_drs_for_gs_storage(uri)
         blob = client.bucket(drs_info.bucket_name, user_project=google_billing_project).get_blob(drs_info.key)
         return VCFInfo.with_blob(blob)
-    if uri.startswith("gs://"):
+    elif uri.startswith("gs://"):
         bucket, key = uri[5:].split("/", 1)
         blob = gs.get_client().bucket(bucket, user_project=google_billing_project).get_blob(key)
         return VCFInfo.with_blob(blob)
+    elif uri.startswith("s3://"):
+        raise ValueError("S3 URIs not supported")
     else:
         return VCFInfo.with_file(uri)
 
--- a/terra_notebook_utils/vcf.py
+++ b/terra_notebook_utils/vcf.py
@@ -2,6 +2,7 @@
 VCF file utilities
 """
 import io
+import gzip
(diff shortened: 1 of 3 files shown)


terra_notebook_utils/vcf.py
```diff
@@ -88,6 +87,18 @@ def with_file(cls, filepath, read_buf: memoryview=None):
...
+    if uri.startswith("drs://"):
+        client, drs_info = drs.resolve_drs_for_gs_storage(uri)
+        blob = client.bucket(drs_info.bucket_name, user_project=google_billing_project).get_blob(drs_info.key)
+        return VCFInfo.with_blob(blob)
+    if uri.startswith("gs://"):
+        bucket, key = uri[5:].split("/", 1)
+        blob = gs.get_client().bucket(bucket, user_project=google_billing_project).get_blob(key)
+        return VCFInfo.with_blob(blob)
+    else:
+        return VCFInfo.with_file(uri)
```
DailyDreaming, line 101: Should this check (and potentially fail with `NotImplementedError`) for non-local schema like `s3://`?

tests/infra/__init__.py
```diff
@@ -15,7 +16,15 @@ def setUp(self):
...
+    sys.stdout = io.TextIOWrapper(io.BytesIO(), sys.stdout.encoding)
     yield
     sys.stdout.close()
     sys.stdout = old_stdout
+
+def upload_data(uri: str, data: bytes):
+    if uri.startswith("gs://"):
+        bucket, key = uri[5:].split("/", 1)
+        Client().bucket(bucket).blob(key).upload_from_file(io.BytesIO(data))
+    else:
```
DailyDreaming, line 28: Same as above, though possibly checking for `drs://` if this is never expected to run `drs` (and maybe a comment about why no drs here).

DailyDreaming (approved):
LGTM.  See comments.

terra_notebook_utils/vcf.py
xbrianh: Will add

tests/infra/__init__.py
xbrianh: This is not a user space function, but I'll add a "not supprted".

Devs attempting to upload to drs uris have deep misconceptions.

xbrianh: On second thought, I'm going to leave this as-is. Adding arbitrary checks for hypothetical developer misconceptions doesn't seem worth the complexity.

PR #203, opened 2020-11-06
Refactor gs tests into separate file

xbrianh:
depends on #196 #198 #199 #200 #201 #202


DailyDreaming (approved):
LGTM, though I'd intuitively expect a file called `test_gs` to have more tests than this.

PR #204, opened 2020-11-06
Parallelizable tests

xbrianh:
depends on #196 #198 #199 #200 #201 #202 #203

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 include common.mk
 
 MODULES=terra_notebook_utils tests
+tests:=$(wildcard tests/test_*.py)
 
 export TNU_TESTMODE?=workspace_access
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,9 +22,13 @@ lint:
...
 
-tests:
-	PYTHONWARNINGS=ignore:ResourceWarning coverage run --source=terra_notebook_utils \
-		-m unittest discover --start-directory tests --top-level-directory . --verbose
+test: $(tests)
+	coverage combine
+	rm -f .coverage.*
+
+# A pattern rule that runs a single test script
+$(tests): %.py : mypy lint
```
DailyDreaming, line 30: Seems like mypy and flake8 are going to be run for the entire codebase for each test when `make test` is run?  Could this be refactored?

xbrianh: They do not run for each test.

You can verify by looking for flake8/mypy output from `make  test`

Issue #207, opened 2020-11-06
table.fetch_drs_url fails for data tables with the "pfb" attribute namespace


mikebaumann:
**Current Behavior**
Using the very useful and important `table.fetch_drs_url `API function fails for data tables with attributes with the "pfb" attribute namespace, as all data handed off after 10/21/20 does.
This was observed with tnu version 0.6.1.
The error is shown below:
```
drs_url = table.fetch_drs_url("submitted_aligned_reads", "NA12842.final.cram")

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-7-d3d4574260bc> in <module>
----> 1 drs_url = table.fetch_drs_url("submitted_aligned_reads", "NA12842.final.cram")

~/notebooks/packages/terra_notebook_utils/table.py in fetch_drs_url(table, file_name)
     41 
     42 def fetch_drs_url(table: str, file_name: str):
---> 43     val = fetch_object_id(table, file_name)
     44     if not val.startswith("drs://"):
     45         raise ValueError(f"Expected DRS url in {table} for {file_name}, got {val} instead.")

~/notebooks/packages/terra_notebook_utils/table.py in fetch_object_id(table, file_name)
     38     DRS urls, when available, are stored in `object_id`.
     39     """
---> 40     return fetch_attribute(table, "file_name", file_name, "object_id")
     41 
     42 def fetch_drs_url(table: str, file_name: str):

~/notebooks/packages/terra_notebook_utils/table.py in fetch_attribute(table, filter_column, filter_val, attribute)
     28     """
     29     for item in _iter_table(table):
---> 30         if filter_val == _get_item_val(item, filter_column):
     31             return _get_item_val(item, attribute)
     32     else:

~/notebooks/packages/terra_notebook_utils/table.py in _get_item_val(item, key)
     21         return item['name']
     22     else:
---> 23         return item['attributes'][key]
     24 
     25 def fetch_attribute(table: str, filter_column: str, filter_val: str, attribute: str):

KeyError: 'file_name'
```

**Expected Behavior**
The `table.fetch_drs_url` function should work as intended both for both data table attributes with and without the "pfb" attribute namespace. Continuing to work for data tables without the namespace is important for earlier users (e.g., Fellows Cohort 1) that have already handed-off their data to their workspace (without the "pfb" namespace) yet want/need to use the latest version of tnu.

Thank you!

PR #216, opened 2020-11-09
Refactor table fixture generation

xbrianh:
This makes it easier to format and add table fixtures

related to #207


DailyDreaming (approved):
LGTM.

PR #217, opened 2020-11-09
Update `fetch_drs_url` to work with new pfb format

### LINKED ISSUE #207: table.fetch_drs_url fails for data tables with the "pfb" attribute namespace

xbrianh:
Support Gen3->Terra PFB export column headers, which are prefixed with "pfb:" as of October 21, 2020.

depends on  #216 
fixes #207

### DIFF
--- a/tests/test_table.py
+++ b/tests/test_table.py
@@ -21,18 +21,24 @@ class TestTerraNotebookUtilsTable(SuppressWarningsMixin, unittest.TestCase):
     @testmode("workspace_access")
     def test_fetch_attribute(self):
         table_name = "simple_germline_variation"
-        filter_column = "md5sum"
+        filter_column = "pfb:md5sum"
         filter_val = "d20e5f752a0d55f0a360b7abe1c8499d"
-        key = "object_id"
+        key = "pfb:object_id"
         val = table.fetch_attribute(table_name, filter_column, filter_val, key)
         self.assertEqual(val, "drs://dg.4503/6e73a376-f7fd-47ed-ac99-0567bb5a5993")
 
     @testmode("workspace_access")
     def test_fetch_object_id(self):
-        table_name = "simple_germline_variation"
-        file_name = "NWD531899.freeze5.v1.vcf.gz"
-        val = table.fetch_object_id(table_name, file_name)
-        self.assertEqual(val, "drs://dg.4503/651a4ad1-06b5-4534-bb2c-1f8ed51134f6")
+        with self.subTest("new pfb format (column headers prefixed with 'pfb:')"):
+            table_name = "simple_germline_variation"
+            file_name = "NWD531899.freeze5.v1.vcf.gz"
+            val = table.fetch_object_id(table_name, file_name)
+            self.assertEqual(val, "drs://dg.4503/651a4ad1-06b5-4534-bb2c-1f8ed51134f6")
+        with self.subTest("old format"):
+            table_name = "simple_germline_variation_old_format"
+            file_name = "NWD531899.freeze5.v1.vcf.gz"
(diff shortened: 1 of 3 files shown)


tests/test_table.py
```diff
@@ -21,18 +21,24 @@ class TestTerraNotebookUtilsTable(SuppressWarningsMixin, unittest.TestCase):
     @testmode("workspace_access")
     def test_fetch_attribute(self):
         table_name = "simple_germline_variation"
-        filter_column = "md5sum"
+        filter_column = "pfb:md5sum"
```
DailyDreaming, line 24: Should these tests be changed to continue to test the old prefix, for example, using something like `for filter_column in ["pfb:md5sum", "md5sum"]`?

tests/test_table.py
```diff
@@ -41,7 +47,7 @@ def test_get_access_token(self):
     @testmode("workspace_access")
     def test_print_column(self):
         table_name = "simple_germline_variation"
-        column = "file_name"
+        column = "pfb:file_name"
```
DailyDreaming, line 50: Similar to the above, should these tests be changed to continue to test the old prefix, for example, using something like `for column in ["pfb:file_name", "file_name"]`?

DailyDreaming (approved):
LGTM.  See comments.

tests/test_table.py
xbrianh: The "pfb:" prefix is incidental to the fixture and not related to tests.

PR #218, opened 2020-11-09
Add `fetch_drs_url` CLI command

### LINKED ISSUE #208: TNU provides a table.fetch_drs_url API function yet not a corresponding CLI command
**Current Behavior**
TNU provides a table.fetch_drs_url API function yet not a corresponding CLI command.
This was observed in tnu version 0.6.1.

**Expected Behavior**
It would be nice if TNU provided a `table.fetch_drs_url` CLI command.
Many of our Fellows come from HPC environments and are familiar with bash scripting, and this CLI command would be useful for enabling that. And, for one-off interactive use from the Terminal, etc.
I see this as a medium/low priority, a "nice to have".

xbrianh:
depends on #216 #217 
closes #208

### DIFF
--- a/terra_notebook_utils/cli/table.py
+++ b/terra_notebook_utils/cli/table.py
@@ -60,6 +60,17 @@ def get_row(args: argparse.Namespace):
     data[f'{args.table}_id'] = e['name']
     print(json.dumps(data, indent=2))
 
+@table_cli.command("fetch-drs-url", arguments={
+    "--table": dict(type=str, required=True, help="table name"),
+    "--file-name": dict(type=str, required=True, help="file name"),
+})
+def fetch_drs_url(args: argparse.Namespace):
+    """
+    Fetch the DRS URL associated with `--file-name` in `--table`.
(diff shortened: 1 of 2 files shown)


terra_notebook_utils/cli/table.py
```diff
@@ -60,6 +60,17 @@ def get_row(args: argparse.Namespace):
     data[f'{args.table}_id'] = e['name']
     print(json.dumps(data, indent=2))
 
+@table_cli.command("fetch-drs-url", arguments={
+    "--table": dict(type=str, required=True, help="table name"),
```
DailyDreaming, line 64: Is it possible to fetch the drs uri without the table name?

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/cli/table.py
xbrianh: That would involve assumptions about table content that could be violated by user actions.

PR #223, opened 2020-11-09
Use basename of source uri, not drs uri

xbrianh:
DRS URIs are opaque. It is better to use basename of cloud URI.


DailyDreaming (approved):
LGTM.  I definitely find this to be an improvement tbh.

PR #238, opened 2020-12-15
Improve cost estimator code quality

xbrianh:


terra_notebook_utils/workflows.py
```diff
@@ -100,8 +98,8 @@ def _parse_machine_type(execution_metadata: dict) -> Tuple[int, int]:
         raise TNUCostException(f"Cannot estimate costs for machine type '{machine_type}'"
                                "Please contact terra-notebook-utils maintainers to add support")
     try:
-        cpus, memory_mb = int(parts[1]), int(parts[2])
-        return cpus, memory_mb
+        cpus, memory_gb = int(parts[1]), int(int(parts[2]) / 1024)
```
aofarrel, line 101: How will this handle situations where the memory is a float? For instance, in the old cost estimator, 6.5 GB got rounded up to 7 GB when reported to the user.

xbrianh: Good point, will switch to floats.

DailyDreaming (approved):
LGTM.

PR #239, opened 2020-12-15
Use JMESPath to extract cost metadata

xbrianh:
depends on #238

### DIFF
--- a/terra_notebook_utils/workflows.py
+++ b/terra_notebook_utils/workflows.py
@@ -54,7 +54,7 @@ def get_workflow(submission_id: str,
 def _get(path: str, data: Dict[str, Any]) -> Any:
     res = jmespath.search(path, data)
     if res is None:
-        raise TNUCostException(f"'{path}' note found in {json.dumps(data)}")
+        raise TNUCostException(f"'{path}' not found in {json.dumps(data)}")
     return res
 
 def estimate_workflow_cost(submission_id: str,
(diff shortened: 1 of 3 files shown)


terra_notebook_utils/workflows.py
```diff
@@ -50,6 +51,12 @@ def get_workflow(submission_id: str,
     resp.raise_for_status()
     return resp.json()
 
+def _get(path: str, data: Dict[str, Any]) -> Any:
+    res = jmespath.search(path, data)
+    if res is None:
+        raise TNUCostException(f"'{path}' note found in {json.dumps(data)}")
```
aofarrel, line 57: If I understand this correctly, it should be:

raise TNUCostException(f"'{path}' **not** found in {json.dumps(data)}")

other than that LGTM

xbrianh: Thanks for catching that.

terra_notebook_utils/workflows.py
```diff
@@ -50,6 +51,12 @@ def get_workflow(submission_id: str,
     resp.raise_for_status()
     return resp.json()
 
+def _get(path: str, data: Dict[str, Any]) -> Any:
+    res = jmespath.search(path, data)
+    if res is None:
+        raise TNUCostException(f"'{path}' not found in {json.dumps(data)}")
```
DailyDreaming, line 57: Consider printing with `indent=4` or `indent=2`.

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/workflows.py
xbrianh: will change

PR #240, opened 2020-12-15
Reorganize cost metadata parsing

xbrianh:
depends on #238 #239

PR #241, opened 2020-12-15
Improve cost reporting for call-cached workflows.

xbrianh:
depends on #238 #239 #240

### DIFF
--- a/terra_notebook_utils/cli/workflows.py
+++ b/terra_notebook_utils/cli/workflows.py
@@ -79,6 +79,7 @@ def estimate_submission_cost(args: argparse.Namespace):
                           ("cpus", 5),
                           ("memory (GB)", 12),
                           ("duration (h)", 13),
+                          ("call cached", 12),
                           ("cost", 5)])
     reporter.print_headers()
     total = 0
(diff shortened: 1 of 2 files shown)


aofarrel:
It seems this doesn't tell users that a workflow is call cached. Doesn't that mean if a user has a history like this where a workflow cost $100 per input file n (nightmareish but work with me here)

workflowA --> foo on [A,B,C] --> bar on file [A,B,C] --> done, costs $300.00
workflowB --> foo on [A,D,E] --> bar on file [A,D,E] --> done, costs $200.01 because A is call-cached

...would get incorrectly reported as $600?

How about the case where workflow C and workflow D are just copies of each other, ie, one cost $100 and the other a penny?

xbrianh:
@aofarrel It _does_ report call caching:
```
(ai) green-beret ~/terra-notebook-utils (xbrianh-cost-estimator-call-cache)>scripts/tnu workflows estimate-submission-cost --workspace terra-notebook-utils-tests --submission-id 7d4d4bbd-6d3a-4e8f-848d-3992f5bd8e33
                          workflow_id  shard  cpus  memory (GB)  duration (h)  call cached  cost
 e50e6808-81bf-487e-9388-10e514be24bf      1     0            0          0.00         True  0.00
 e4a31794-7e29-4d3b-ad9a-ad1a2ddc43bc      1     2            2          3.65        False  0.27
------------------------------------------------------------------------------------------------
                                                                                            0.27
```

aofarrel:
I don't know how I missed that lol

LGTM!

terra_notebook_utils/cli/workflows.py
```diff
@@ -122,7 +127,7 @@ def print_divider(self):
         print("-" * self.width)
 
     def ff(self, val: Any, width: int, decimals: int=2) -> str:
```
DailyDreaming, line 129: Should `val: Any` be `Union[str, bool, int, float]`?

DailyDreaming (approved):
LGTM.  See comment.  Also... I know floats in python can be printed to be fairly long.  Consider a a short width and appending a unit (and potentially converting if it gets too long).

terra_notebook_utils/cli/workflows.py
xbrianh: Probably, but that should be handled outside of this PR :)

xbrianh:
@DailyDreaming As shown in the output pasted in above, the floats are constrained to two decimal points. This is handled in the function you commented on, in fact.

DailyDreaming:
Understood.  I should have made it clearer that I was asking.

PR #243, opened 2021-01-06
Introduce async context manager utility

xbrianh:
This is a convenience class useful for providing asynchronous execution to user
facing context managers.

### DIFF
--- a/terra_notebook_utils/utils.py
+++ b/terra_notebook_utils/utils.py
@@ -1,0 +1,34 @@
+from concurrent.futures import ThreadPoolExecutor, Future, as_completed
+from typing import Any, Callable, Set
+
+
+class _AsyncContextManager:
+    """
+    Context manager for asynchronous execution. Wait on exit for all jobs to complete.
+    """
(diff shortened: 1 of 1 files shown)


terra_notebook_utils/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        for f in self._futures.copy():
+            if f.done():
+                f.result()
+                self._futures.remove(f)
+
+    def _prepare_for_exit(self):
+        raise NotImplementedError()
+
+    def __enter__(self):
+        self._executor = ThreadPoolExecutor(max_workers=10)
```
DailyDreaming, line 24: Should `max_workers` be proportional to max_cpu?

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/utils.py
xbrianh: Nope: this is an IO bound task, so it shouldn’t be set to the number of CPUs. I’m actually not sure what it should be set to. In the meantime, 10 seems like a decent guess according to empirical evidence.

PR #244, opened 2021-01-06
Context manager to upload Terra data tables

xbrianh:
This transparantly handles typed data uploads to Terra. Uploads are
performed in the background.

Tests provided in future PR :)

Depends on #243

### DIFF
--- a/terra_notebook_utils/table.py
+++ b/terra_notebook_utils/table.py
@@ -42,7 +42,7 @@ class Writer(_AsyncContextManager):
         request_data = list()
         for name, val in row.attributes.items():
             if isinstance(val, str):
-                update_ops = list()
+                update_ops = list()  # No Firecloud update operations needed for string values
             elif isinstance(val, (int, float, bool)):
                 update_ops = [dict(op="AddUpdateAttribute", attributeName=name, addUpdateAttribute=val)]
             elif hasattr(val, "__iter__"):
@@ -52,10 +52,6 @@ class Writer(_AsyncContextManager):
                     update_ops.append(dict(op="AddListMember", attributeListName=name, newMember=m))
                     types.add(type(m))
                 assert 1 == len(types)
-# assert types.pop() in VALUE.__args__, f"Row: {row.name}: List members must be of type {VALUE.__args__}"
-            else:
-                pass
-                # raise TypeError(f"Row {row.name}: data table value must be of type{VALUE.__args__}")
             request_data.extend(update_ops)
         return request_data
 
--- a/terra_notebook_utils/table.py
+++ b/terra_notebook_utils/table.py
@@ -1,10 +1,131 @@
 """
 Terra data table commands
 """
+import os
+from uuid import uuid4
+from collections import defaultdict, namedtuple
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Union
+
+import requests
 from firecloud import fiss
-from typing import Optional
 
 from terra_notebook_utils import WORKSPACE_GOOGLE_PROJECT, WORKSPACE_NAME
+from terra_notebook_utils.utils import _AsyncContextManager
+
+
+Row = namedtuple("Row", "name attributes")
+
+VALUE = Union[str, int, float, bool]
+ATTRIBUTES = Dict[str, Union[VALUE, Iterable[VALUE]]]
+UPDATE_OPS = List[Dict[str, Any]]
+COLUMN_HEADERS = Tuple[str, ...]
+ROW_LIKE = Tuple[str, ATTRIBUTES]
+
+class Writer(_AsyncContextManager):
+    """
+    Distribute row uploads across as few API calls as possible.
+    Uploads are performed in the background.
+    Also transparantly handles sequences, which the Firecloud API makes difficult. (These must be uploaded and modified
+    via separate API calls.)
+    """
+    def __init__(self, name: str, **kwargs):
+        self.name = name
+        self._init_request_data()
+        self._workspace_name = kwargs.get("workspace_name", WORKSPACE_NAME)
+        self._workspace_google_project = kwargs.get("workspace_google_project", WORKSPACE_GOOGLE_PROJECT)
+
+    def _init_request_data(self):
+        self._tsvs: Dict[COLUMN_HEADERS, str] = defaultdict(list)
+        self._row_update_request_data: Dict[COLUMN_HEADERS, List[Tuple[Row, UPDATE_OPS]]] = dict()
+
+    def _get_row_update_request_data(self, row: Row) -> List[Dict[str, Any]]:
+        request_data = list()
+        for name, val in row.attributes.items():
+            if isinstance(val, str):
+                update_ops = list()
+            elif isinstance(val, (int, float, bool)):
+                update_ops = [dict(op="AddUpdateAttribute", attributeName=name, addUpdateAttribute=val)]
+            elif hasattr(val, "__iter__"):
+                update_ops = [dict(op="RemoveAttribute", attributeName=name)]
+                types: Set[type] = set()
+                for m in val:
+                    update_ops.append(dict(op="AddListMember", attributeListName=name, newMember=m))
+                    types.add(type(m))
+                assert 1 == len(types)
+# assert types.pop() in VALUE.__args__, f"Row: {row.name}: List members must be of type {VALUE.__args__}"
+            else:
+                pass
+                # raise TypeError(f"Row {row.name}: data table value must be of type{VALUE.__args__}")
+            request_data.extend(update_ops)
+        return request_data
+
+    def put_row(self, item: Union[ROW_LIKE, ATTRIBUTES]) -> str:
+        if isinstance(item, dict):
+            row = Row(f"{uuid4()}", item)
+        else:
+            row = Row(*item)
+        column_headers = tuple(sorted(row.attributes.keys()))
+        if column_headers not in self._tsvs:
+            self._tsvs[column_headers] = "\t".join([f"{self.name}_id", *column_headers])
+            self._row_update_request_data[column_headers] = list()
+        self._tsvs[column_headers] += (
+            os.linesep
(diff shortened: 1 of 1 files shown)


terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+            elif isinstance(val, (int, float, bool)):
+                update_ops = [dict(op="AddUpdateAttribute", attributeName=name, addUpdateAttribute=val)]
+            elif hasattr(val, "__iter__"):
+                update_ops = [dict(op="RemoveAttribute", attributeName=name)]
+                types: Set[type] = set()
+                for m in val:
+                    update_ops.append(dict(op="AddListMember", attributeListName=name, newMember=m))
+                    types.add(type(m))
+                assert 1 == len(types)
+# assert types.pop() in VALUE.__args__, f"Row: {row.name}: List members must be of type {VALUE.__args__}"
```
DailyDreaming, line 55: Leftovers?

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+            else:
+                pass
+                # raise TypeError(f"Row {row.name}: data table value must be of type{VALUE.__args__}")
```
DailyDreaming, line 58: More leftovers?

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+        else:
+            row = Row(*item)
+        column_headers = tuple(sorted(row.attributes.keys()))
+        if column_headers not in self._tsvs:
+            self._tsvs[column_headers] = "\t".join([f"{self.name}_id", *column_headers])
+            self._row_update_request_data[column_headers] = list()
+        self._tsvs[column_headers] += (
+            os.linesep
+            + row.name + "\t"
+            + "\t".join(row.attributes[c] if isinstance(row.attributes[c], str)
```
DailyDreaming, line 74: What is the `x` default?  Is it possible this could be an `int` or is this only handling the `None` case?

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+        except requests.exceptions.HTTPError as e:
+            if 500 == e.response.status_code:
+                # Firecloud occasionally throws 500 errors for succesful update operations.
+                # Check if we get the row we expect, retry otherwise.
+                if 5 > retry:
+                    cur_row = get_row(self.name, row.name)
+                    if row != cur_row:
+                        self.submit(self._do_fiss_updates, row, request_data, retry=retry + 1)
+                else:
+                    raise Exception("Ran out of retries updating row {row.name}") from e
```
DailyDreaming, line 122: Missing `f`.

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+                                    row.name,
+                                    request_data).raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            if 500 == e.response.status_code:
+                # Firecloud occasionally throws 500 errors for succesful update operations.
+                # Check if we get the row we expect, retry otherwise.
+                if 5 > retry:
+                    cur_row = get_row(self.name, row.name)
+                    if row != cur_row:
+                        self.submit(self._do_fiss_updates, row, request_data, retry=retry + 1)
```
DailyDreaming, line 120: Should this be incrementing a retry every time instead of subtracting?  I'm not very clear on how this is retrying.

Possibly consider a way to make this work with a `retry` decorator.

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+
+    def _do_fiss_updates(self, row: Row, request_data: UPDATE_OPS, retry: int=0):
+        try:
+            fiss.fapi.update_entity(self._workspace_google_project,
+                                    self._workspace_name,
+                                    self.name,
+                                    row.name,
+                                    request_data).raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            if 500 == e.response.status_code:
```
DailyDreaming, line 114: Would `if e.response.status_code >= 500:` or `if e.response.status_code in (500, 502, 503, 504):` make sense here?

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+            elif hasattr(val, "__iter__"):
+                update_ops = [dict(op="RemoveAttribute", attributeName=name)]
+                types: Set[type] = set()
+                for m in val:
+                    update_ops.append(dict(op="AddListMember", attributeListName=name, newMember=m))
+                    types.add(type(m))
+                assert 1 == len(types)
+# assert types.pop() in VALUE.__args__, f"Row: {row.name}: List members must be of type {VALUE.__args__}"
+            else:
+                pass
```
DailyDreaming, line 57: Nit: no op

```
else:
    pass
```

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+                        else "x"
+                        for c in column_headers)
+        )
+        update_request_data = self._get_row_update_request_data(row)
+        if update_request_data:
+            self._row_update_request_data[column_headers].append((row, update_request_data))
+        self._upload(1024 * 20)
+        return row.name
+
+    def _upload(self, threshold: Optional[int]=None):
```
DailyDreaming, line 84: Type hint: ` -> None`

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+    def _do_fiss_upload(self, tsv: str, row_update_request_data: List[Tuple[Row, List[Dict[str, Any]]]]):
+        fiss.fapi.upload_entities(self._workspace_google_project,
+                                  self._workspace_name,
+                                  tsv,
+                                  model="flexible").raise_for_status()
+        for row, request_data in row_update_request_data:
+            if request_data:
+                self.submit(self._do_fiss_updates, row, request_data)
+
+    def _do_fiss_updates(self, row: Row, request_data: UPDATE_OPS, retry: int=0):
```
DailyDreaming, line 106: Type hint: ` -> None`

DailyDreaming (changes requested):
Most are minor comments.   I'm concerned about one of the retries potentially going infinite.

DailyDreaming (approved):

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+    Also transparantly handles sequences, which the Firecloud API makes difficult. (These must be uploaded and modified
+    via separate API calls.)
+    """
+    def __init__(self, name: str, **kwargs):
+        self.name = name
+        self._init_request_data()
+        self._workspace_name = kwargs.get("workspace_name", WORKSPACE_NAME)
+        self._workspace_google_project = kwargs.get("workspace_google_project", WORKSPACE_GOOGLE_PROJECT)
+
+    def _init_request_data(self):
```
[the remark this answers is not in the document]
xbrianh: I’m happy to omit type hints for methods without a return value.

Do you think we should make it explicit?

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+from terra_notebook_utils.utils import _AsyncContextManager
+
+
+Row = namedtuple("Row", "name attributes")
+
+VALUE = Union[str, int, float, bool]
+ATTRIBUTES = Dict[str, Union[VALUE, Iterable[VALUE]]]
+UPDATE_OPS = List[Dict[str, Any]]
+COLUMN_HEADERS = Tuple[str, ...]
+ROW_LIKE = Tuple[str, ATTRIBUTES]
```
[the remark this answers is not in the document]
xbrianh: These types are only used in this file, so I think it’s simpler to keep them here for now.

terra_notebook_utils/table.py
```diff
@@ -1,10 +1,131 @@
...
+
+    def _init_request_data(self):
+        self._tsvs: Dict[COLUMN_HEADERS, str] = defaultdict(list)
+        self._row_update_request_data: Dict[COLUMN_HEADERS, List[Tuple[Row, UPDATE_OPS]]] = dict()
+
+    def _get_row_update_request_data(self, row: Row) -> List[Dict[str, Any]]:
+        request_data = list()
+        for name, val in row.attributes.items():
+            if isinstance(val, str):
+                update_ops = list()
```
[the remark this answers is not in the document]
xbrianh: It’s an empty list because of Firecloud esoterica. Will add comment.

terra_notebook_utils/table.py
xbrianh: Yep, thanks!

terra_notebook_utils/table.py
xbrianh: Lost in the refrigerator. Will fix.

terra_notebook_utils/table.py
xbrianh: Not an odd choice to me :)

If this pattern is used more than once we can consider an abstraction.

terra_notebook_utils/table.py
xbrianh: Possibly. But without knowledge of what `50x` errors means, it’s difficult to know what to do.

For instance, this line handles the possibility of a successful request returning `500`. This is pathological behavior from Firecould. What should we do with `50x` errors? Who knows.

So you may be right. However, I’d prefer to handle the error narrowly until there’s evidence of pathological `50x` errors.

terra_notebook_utils/table.py
xbrianh: Will remove.

terra_notebook_utils/table.py
xbrianh: Will fix

terra_notebook_utils/table.py
xbrianh: Only string values can be uploaded in the original TSV upload. Other types require additional Firecloud API calls. “x” is a placeholder value.

Will add comment.

PR #253, opened 2021-01-11
Refactor table deletion

xbrianh:


terra_notebook_utils/table.py
```diff
@@ -231,21 +231,17 @@ def get_row(table: str,
...
-def delete_table(ent_type: str):
-    delete_entities([e for e in list_entities(ent_type)])
-
 def upload_entities(tsv_data,
                     workspace_name: Optional[str]=WORKSPACE_NAME,
                     workspace_google_project: Optional[str]=WORKSPACE_GOOGLE_PROJECT):
     resp = fiss.fapi.upload_entities(workspace_google_project, workspace_name, tsv_data, model="flexible")
     resp.raise_for_status()
 
+def delete(table: str, **kwargs):
```
DailyDreaming, line 240: Nit: Grab bag `**kwargs` are kind of unfriendly.  It avoids typing and documentation about what inputs are expected.  Consider putting this information into the function docstring if it's unavoidable.

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/table.py
xbrianh: This is a fair point about readability, and I shared this concern during implementation.

I will likely revert to the original explicit arguments, but the work will be deferred to future PRs. This reply should suffice against the multiple comments complaining about **kwargs :)

PR #255, opened 2021-01-11
Refactor table.get_row

xbrianh:
depends on #254

### DIFF
--- a/terra_notebook_utils/table.py
+++ b/terra_notebook_utils/table.py
@@ -20,6 +20,7 @@ ATTRIBUTES = Dict[str, Union[VALUE, Iterable[VALUE]]]
 UPDATE_OPS = List[Dict[str, Any]]
 COLUMN_HEADERS = Tuple[str, ...]
 ROW_LIKE = Tuple[str, ATTRIBUTES]
+ROW_OR_NAME = Union[Row, str]
 
 # A note on types:
 # As presented in Terra's UI, tables may contain string, integer, and boolean values.  However, Firecloud API clients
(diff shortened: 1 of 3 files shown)


terra_notebook_utils/table.py
```diff
@@ -215,13 +216,21 @@ def list_rows(table: str, **kwargs) -> Generator[Row, None, None]:
...
         page_number += 1
 
-def get_row(table: str,
-            entity_id: str,
-            workspace_name: Optional[str]=WORKSPACE_NAME,
-            workspace_google_project: Optional[str]=WORKSPACE_GOOGLE_PROJECT):
-    resp = fiss.fapi.get_entity(workspace_google_project, workspace_name, table, entity_id)
-    resp.raise_for_status()
-    return resp.json()
+def get_row(table: str, item: ROW_OR_NAME, **kwargs) -> Optional[Row]:
```
DailyDreaming, line 219: Echoing an earlier comment:

```
Nit: Grab bag `**kwargs` are kind of unfriendly.  It avoids typing and documentation about what inputs are expected.  Consider putting this information into the function docstring if it's unavoidable.
```

I'm not sure if your style is changing but this is similar to Hannes's style if you take a look at Azul.  This is also something that was throughout Toil when I first joined.  It's not so unclear when used sparingly, but... I have particularly strong memories of how confusing it can be when over-used, especially if lacking docstrings (or worse, the docstrings are outdated and actually document the wrong kwargs).

DailyDreaming (approved):
Approving, but see comment.

PR #256, opened 2021-01-11
Introduce table upload convenience methods

xbrianh:
depends on #253 #254 #255

### DIFF
--- a/terra_notebook_utils/cli/table.py
+++ b/terra_notebook_utils/cli/table.py
@@ -2,6 +2,7 @@ import os
 import json
 import typing
 import argparse
+from uuid import uuid4
 
 from terra_notebook_utils import table as tnu_table
 from terra_notebook_utils.cli import dispatch, Config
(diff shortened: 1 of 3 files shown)


terra_notebook_utils/cli/table.py
```diff
@@ -70,7 +71,7 @@ def fetch_drs_url(args: argparse.Namespace):
 
 @table_cli.command("put-row", arguments={
     "--table": dict(type=str, required=True, help="table name"),
-    "--id": dict(type=str, required=True, help="Row id. This should be unique for the table"),
+    "--row": dict(type=str, required=False, default=None, help="row name"),
```
DailyDreaming, line 74: Do these still need to be unique?  If so, I thought that that was helpful info.  I might also document the behavior if this is not specified.

tests/test_table.py
```diff
@@ -25,19 +25,8 @@ def test_get_access_token(self):
     @testmode("workspace_access")
     def test_table(self):
         table_name = f"test_{uuid4()}"
-        number_of_entities = 5
         table.delete(table_name)  # remove cruft from previous failed tests
```
DailyDreaming, line 28: I don't understand this test.  Is this testing the deletion of a non-existent table?

DailyDreaming (approved):
See comments.

tests/test_table.py
xbrianh: These are updated in a separate PR.

terra_notebook_utils/cli/table.py
xbrianh: Fair point. Will update.

PR #257, opened 2021-01-11
Introduce table delete convenience methods

xbrianh:
depends on #256


DailyDreaming (approved):
Approving but earlier comments about `kwargs` still apply.

PR #261, opened 2021-01-12
Improves the readability of cost estimation

xbrianh:
Also introduce time units that will be useful in the future.

### DIFF
--- a/terra_notebook_utils/costs.py
+++ b/terra_notebook_utils/costs.py
@@ -1,27 +1,31 @@
 from math import ceil
 
 
-MINUTE = 60
+SECOND = 1.0
+MINUTE = SECOND * 60
+HOUR = MINUTE * 60
+DAY = HOUR * 24
+MONTH = DAY * 30.437  # Not critical to be exact
 
 class GCPCustomN1Cost:
     # Pricing information: https://cloud.google.com/compute/vm-instance-pricing#n1_custommachinetypepricing
-    on_demand_per_hour_per_cpu = 0.033174
-    on_demand_per_hour_per_gb = 0.004446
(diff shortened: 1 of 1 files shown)


aofarrel (changes requested):
My gut instinct is to use [time](https://docs.python.org/3/library/time.html) whenever I touch anything greater than minutes but if this is all you're going to do with dates then I can't fault it, we don't need leap day handling and we're not parsing much in the way of specific calendar dates.

That being said, I don't think month is a useful display. The length of a month is not constant, so it loses usefulness. "42 days" is more useful than "1-point-who-knows-what months" even if we tell the user a month is X number of days. It also can confuse people because they may see a month as something static -- ie, they might interpret it as "next month" instead of "30ish days from now" because we tend to think of months as static blocks that divide the year rather than a measurement of time between two events.

TL;DR I argue removing months on the grounds of human psychology and [Julias Ceaser's ambitions](https://en.wikipedia.org/wiki/Julian_calendar).

aofarrel:
That's what I mean. The readability of "1.43 months" is less than "42 days," whether it's directly user-facing or not, right?

DailyDreaming (approved):
LGTM.

xbrianh:
@aofarrel I think `foo = some_cost / MONTH` is more readable that `foo = some_cost / (30.437 * DAY)`. Also, the GCP resource I’m trying to cost out is quoted in monthly units. It’s likely you pay monthly rent, every.

In any case, this is probably too much time spent on a nit.

aofarrel:
> Also, the GCP resource I’m trying to cost out is quoted in monthly units. It’s likely you pay monthly rent, every.

oops

I getcha now! Makes sense!

PR #263, opened 2021-01-13
Don't attempt to estimate subworkflow costs

xbrianh:
These will be handled separately in future PR.

depends on #262

PR #264, opened 2021-01-13
Method to get all (sub-)workflows for a submission.

xbrianh:
depends on #263

### DIFF
--- a/terra_notebook_utils/workflows.py
+++ b/terra_notebook_utils/workflows.py
@@ -12,6 +12,7 @@ import jmespath
 from firecloud import fiss
 
 from terra_notebook_utils import WORKSPACE_NAME, WORKSPACE_GOOGLE_PROJECT, costs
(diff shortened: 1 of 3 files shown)


terra_notebook_utils/workflows.py
```diff
@@ -57,6 +57,26 @@ def _get(path: str, data: Dict[str, Any]) -> Any:
         raise TNUCostException(f"'{path}' not found in {json.dumps(data, indent=2)}")
     return res
 
+def get_all_workflows(submission_id: str,
```
DailyDreaming, line 60: This looks like it needs a test added.

DailyDreaming (approved):
Approving but consider the comment.

terra_notebook_utils/workflows.py
xbrianh: simple test added

PR #265, opened 2021-01-13
Include subworkflows in cost estimates

xbrianh:
depends on #264

### DIFF
--- a/tests/test_workflows.py
+++ b/tests/test_workflows.py
@@ -66,7 +66,7 @@ class TestTerraNotebookUtilsWorkflowsCLI(CLITestMixin, unittest.TestCase):
 
     def test_estimate_workflow_cost(self):
         ret = dict(workflows=[dict(workflowId=1)])
-        with mock.patch("terra_notebook_utils.workflows.get_submission", return_value=ret):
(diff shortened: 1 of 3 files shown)


tests/test_workflows.py
```diff
@@ -63,7 +63,7 @@ def test_get_workflow(self):
 
     def test_estimate_workflow_cost(self):
         ret = dict(workflows=[dict(workflowId=1)])
-        with mock.patch("terra_notebook_utils.workflows.get_submission", return_value=ret):
+        with mock.patch("terra_notebook_utils.workflows.get_all_workflows", return_value=ret):
```
DailyDreaming, line 66: `terra_notebook_utils.workflows.get_submission` seems to still exist: https://github.com/DataBiosphere/terra-notebook-utils/blob/xbrianh-subworkflow-costs/terra_notebook_utils/workflows.py#L32

Should it still have a test?

DailyDreaming (approved):
Approving but consider comment.

tests/test_workflows.py
xbrianh: Yes. And it does! The test was not removed.

PR #266, opened 2021-01-14
Introduce function to perform concurrent recursion

xbrianh:
This is useful for IO functions that should be called on their own results,
where order doesn't matter.

### DIFF
--- a/terra_notebook_utils/utils.py
+++ b/terra_notebook_utils/utils.py
@@ -33,13 +33,13 @@ class _AsyncContextManager:
             self.prune_futures()
         self._executor.shutdown()
 
-def concurrent_recursion(process_item: Callable[[Any], Iterable[Any]], items: Iterable[Any], concurrency: int=8):
+def concurrent_recursion(recurse: Callable[[Any], Iterable[Any]], initial_data: Iterable[Any], concurrency: int=8):
     """
-    Call `process_item` on each item in `items`, and on each item returned by `process_item`, concurrently.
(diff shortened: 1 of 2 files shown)


terra_notebook_utils/utils.py
```diff
@@ -32,3 +32,14 @@ def __exit__(self, *args, **kwargs):
                 pass
             self.prune_futures()
         self._executor.shutdown()
+
+def concurrent_recursion(process_item: Callable[[Any], Iterable[Any]], items: Iterable[Any], concurrency: int=8):
```
DailyDreaming, line 36: Nit: ` -> None`

DailyDreaming, line 36: Nit: Two points:

1. I would expect a function with the name `recursion` to be a function calling itself and `concurrent_recursion` does not call `concurrent_recursion`.
2. This may be arguable, but I view recursion is an implementation detail.  Naming that describes the function's purpose rather than how it's implemented may be clearer and more robust against implementation changes in the future.

DailyDreaming (approved):
LGTM.  See comments.

terra_notebook_utils/utils.py
xbrianh: I’ve never used this anywhere and don’t intend to start.

xbrianh: I thought the functions _was_ named for it’s purpose (recursively calling `process_item`).

I’ll see if I can think of better names.

PR #268, opened 2021-01-15
Refactor JMESPath getter, move to utils

xbrianh:
depends on #267


Richard-Hansen (approved):
Looks good.

To be nit-picky, you could test the ```default``` parameter of ```js_get( ... )```, but that's overkill.

xbrianh:
@Richard-Hansen nah, that’s reasonable. Will add test.

PR #271, opened 2021-01-21
Make table API call chunk size configurable

xbrianh:


terra_notebook_utils/table.py
```diff
@@ -90,16 +92,16 @@ def put_row(self, item: Union[ROW_LIKE, ATTRIBUTES]) -> str:
...
+        self._upload()
         return row.name
 
-    def _upload(self, threshold: Optional[int]=None):
+    def _upload(self, force: bool=False):
         """
-        Schedule uploads for all TSVs of size equal to or greater than `threshold`.
-        If `threshold` is None, schedule uploads for all TSVs.
+        Schedule uploads for all TSVs of size equal to or greater than `self._tsv_upload_size`.
+        If `force` is None, schedule uploads for all TSVs.
```
Richard-Hansen, line 101: ```If `force` is True,```

Richard-Hansen (approved):
Small documentation comment.

terra_notebook_utils/table.py
xbrianh: fixed, thanks!

PR #274, opened 2021-02-10
Support table cells with dict values

xbrianh:
depends on #273

### DIFF
--- a/terra_notebook_utils/table.py
+++ b/terra_notebook_utils/table.py
@@ -4,7 +4,7 @@ Terra data table commands
 import os
 from uuid import uuid4
 from collections import defaultdict, namedtuple
-from typing import Any, Dict, Generator, Iterable, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, Generator, Iterable, List, Mapping, Optional, Set, Tuple, Union
 
 import requests
 from firecloud import fiss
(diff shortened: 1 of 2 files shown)


terra_notebook_utils/table.py
```diff
@@ -62,14 +62,19 @@ def _init_request_data(self):
...
         request_data = list()
+        types: Set[type]
         for name, val in row.attributes.items():
             if isinstance(val, str):
                 update_ops = list()  # No Firecloud update operations needed for string values
             elif isinstance(val, (int, float, bool)):
                 update_ops = [dict(op="AddUpdateAttribute", attributeName=name, addUpdateAttribute=val)]
+            elif hasattr(val, "keys") and hasattr(val, "items"):  # TODO: better way to detect dict-like objects?
+                types = {type(k) for k in val.keys()}
+                assert 1 == len(types) and str == types.pop()
```
DailyDreaming, line 73: Is `assert types == {str}` equivalent here?

terra_notebook_utils/table.py
```diff
@@ -62,14 +62,19 @@ def _init_request_data(self):
 
     def _get_row_update_request_data(self, row: Row) -> List[Dict[str, Any]]:
         request_data = list()
+        types: Set[type]
         for name, val in row.attributes.items():
             if isinstance(val, str):
                 update_ops = list()  # No Firecloud update operations needed for string values
             elif isinstance(val, (int, float, bool)):
                 update_ops = [dict(op="AddUpdateAttribute", attributeName=name, addUpdateAttribute=val)]
+            elif hasattr(val, "keys") and hasattr(val, "items"):  # TODO: better way to detect dict-like objects?
```
DailyDreaming, line 71: I believe that `collections.Mapping` would probably work here (e.g. `isinstance(val, Mapping)`).  Dictionaries type as `collections.Mapping` and it seems to define both items and keys: https://docs.python.org/3.1/library/collections.html#collections.Mapping

DailyDreaming (approved):
LGTM.  See comments.

terra_notebook_utils/table.py
DailyDreaming: Edit: This is `collections.abc` now apparently: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

terra_notebook_utils/table.py
xbrianh: Yes, they’re equivalent. I’ll go with your suggestion, which looks cleaner.

terra_notebook_utils/table.py
xbrianh: Good suggestion! I went with [typing.Mapping](https://docs.python.org/3/library/typing.html), which points to the collections definition.

PR #277, opened 2021-02-18
Allow table cells with `None` values

xbrianh:
When row attributes contain `None`, do not upload a value for that cell.


juklucas (approved):
lgtm

Thanks for the quick fix!

PR #282, opened 2021-04-01
Fix workspace resolution for VCF CLI commands

xbrianh:


terra_notebook_utils/cli/vcf.py
```diff
@@ -11,6 +11,11 @@
     "path": dict(
         help="local path, gs://, or drs://"
     ),
+    "--workspace": dict(
+        type=str,
+        default=None,
+        help="workspace name. If not provided, the configured CLI workspace will be used"
```
DailyDreaming, line 17: Consider adding a default: `default=Config.info['workspace']`, and including it in the help with either:

`(default: %(default)s)` (add default for only this): https://docs.python.org/3/library/argparse.html#help

`ArgumentDefaultsHelpFormatter` (add defaults universally): https://docs.python.org/3/library/argparse.html#formatter-class

Also... just a nit and something that may not be important... but adding a period to the end and capitalizing "Workspace" in the help message.

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/cli/vcf.py
xbrianh: Defaults are resolved through `terra_notebook_utils.cli.Config.resolve`

Will fix typos.

PR #283, opened 2021-04-27
Introduce http module

xbrianh:
This is useful for retry configuration and connection pooling. Using request.Session may improve performance for the Python API, e.g. usage in notebook environments.

### DIFF
--- a/tests/infra/server.py
+++ b/tests/infra/server.py
@@ -1,0 +1,37 @@
+import threading
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from typing import Tuple
+
+
+class ThreadedLocalServer(threading.Thread):
+    """
+    Run an http server on a separate thread.
+    """
+    def __init__(self, handler_class: BaseHTTPRequestHandler, address: Tuple[str, int]=('', 8000)):
+        super().__init__(daemon=True)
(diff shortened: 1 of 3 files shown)


tests/infra/server.py
```diff
@@ -0,0 +1,37 @@
+import threading
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from typing import Tuple
+
+
+class ThreadedLocalServer(threading.Thread):
+    """
+    Run an http server on a separate thread.
+    """
+    def __init__(self, handler_class: BaseHTTPRequestHandler, address: Tuple[str, int]=('', 8000)):
```
DailyDreaming, line 10: Suggestion, it might be worth grabbing an unused port with something like:

```
import contextlib
import socket

def unused_tcp_port():
    with contextlib.closing(socket.socket()) as sock:
        sock.bind(('127.0.0.1', 0))
        return sock.getsockname()[1]
```

DailyDreaming (approved):
LGTM.  See comment.

tests/infra/server.py
xbrianh: This is a good suggestion, and I intend to implement it soon. For now, punting to a future PR.

PR #284, opened 2021-04-27
Use http module in gs and drs

xbrianh:
Depends on #283

### DIFF
--- a/terra_notebook_utils/drs.py
+++ b/terra_notebook_utils/drs.py
@@ -4,9 +4,7 @@ Utilities for working with DRS objects
 import os
 import sys
 import re
-import json
 import logging
-import requests
 import traceback
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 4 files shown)


terra_notebook_utils/drs.py
```diff
@@ -70,10 +69,12 @@ def fetch_drs_info(drs_url: str) -> dict:
         'authorization': f"Bearer {access_token}",
         'content-type': "application/json"
     }
+    # drs://dg.712C/57a6f0bb-bc4b-4530-a95e-9c2c3398b618
```
DailyDreaming, line 72: Leftovers?

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/drs.py
xbrianh: yep, thanks!

PR #285, opened 2021-04-27
Use urllib3 Retry for table requests to fiss API

xbrianh:
This performs http retries by changing the retry configuration of the requests.Session object provided by the google SDK, as exposed through the [fiss](https://github.com/broadinstitute/fiss) library.

depends on #283 #284


DailyDreaming (approved):
LGTM.

Issue #294, opened 2021-06-05
Add extensible support for cloud storage


xbrianh:
- Terra will eventually support Azure.
- A general DRS client should support cloud copies to major cloud providers.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/TLP-2083)

PR #295, opened 2021-06-05
Introduce blobstore base class

xbrianh:
This introduces an extendible abstraction layer for remote storage. While it is mainly aimed on services such as S3 and GS, it will also be useful for http/s resolved data, as well as local data.

The following interfaces will be supported in the future:
```
url_blob = URLBlob("https://foo/bar")
gs_blob = GSBlob("a-bucket", "some-key")
s3_blob = GSBlob("a-bucket", "some-key")

# download blobs
url_blob.download("my-local-filepath")
s3_blob.download("my-local-filepath")

# cloud copy
copy(url_blob, gs_blob)
copy(gs_blob, s3_blob)

# upload data
copy(local_blob, gs_blob)
```

related to #294

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ bgzip >= 0.3.5, < 0.4
 cli-builder >= 0.1.5, < 0.2
 oauth2client
 jmespath == 0.10.0
+getm >= 0.0.2, <= 0.1.0
--- a/terra_notebook_utils/blobstore/__init__.py
+++ b/terra_notebook_utils/blobstore/__init__.py
@@ -1,0 +1,95 @@
+from collections import namedtuple
+from typing import Any, Generator
+
+from getm import checksum
+
+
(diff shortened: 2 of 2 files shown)


DailyDreaming (approved):
This looks good to me... and while not included in this PR, but in your example:

```
url_blob = URLBlob("https://foo/bar")
gs_blob = GSBlob("a-bucket", "some-key")
s3_blob = GSBlob("a-bucket", "some-key")
```

I'd just like to say that I'd personally prefer just one `Blob` object, and letting the function do the parsing:

```
url_blob = Blob("https://foo/bar")
gs_blob = Blob("gs://a-bucket/some-key")
s3_blob = Blob("s3://a-bucket/some-key")
```

Just a suggestion.

xbrianh:
@DailyDreaming The user experience is intended to be more or less as you describe. However it is programmatically useful to distinguish blob classes.

PR #296, opened 2021-06-05
Introduce gs blobstore

xbrianh:
depends on #295 
related to #294

### DIFF
--- a/terra_notebook_utils/blobstore/gs.py
+++ b/terra_notebook_utils/blobstore/gs.py
@@ -59,8 +59,10 @@ class GSBlob(blobstore.Blob):
         self.billing_project = billing_project
         self.credentials = credentials
 
-    # The next two methods customize pickling behavior
-    # see docs: https://docs.python.org/3/library/pickle.html#object.__getstate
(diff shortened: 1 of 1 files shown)


terra_notebook_utils/blobstore/gs.py
```diff
@@ -0,0 +1,196 @@
...
+        return blob.exists()
+
+    def size(self) -> int:
+        return self._get_native_blob().size
+
+    def cloud_native_checksum(self) -> str:
+        return self._get_native_blob().crc32c
+
+    @property
+    def Hash(self) -> checksum._Hasher:
```
DailyDreaming, line 142: Nit: I'd expect this to be called `Hasher` and as a user, if I call `Blob.hash()` to get the actual RC32C in return.

DailyDreaming (approved):
Previous comments from the `local` PR apply but are not reiterated here.  Otherwise, LGTM.

terra_notebook_utils/blobstore/gs.py
xbrianh: Fair point. This will be handled in a future PR.

PR #297, opened 2021-06-07
Introduce local blobstore

xbrianh:
depends on #295 
related to #294

### DIFF
--- a/terra_notebook_utils/blobstore/local.py
+++ b/terra_notebook_utils/blobstore/local.py
@@ -2,7 +2,7 @@ import os
 import shutil
 from math import ceil
 from functools import wraps
-from typing import Dict, Generator, Optional
+from typing import Generator
 
 from getm import default_chunk_size
 
@@ -42,6 +42,7 @@ class LocalBlob(blobstore.Blob):
         self._path = os.path.join(basepath, relpath)
 
     # The next two methods customize pickling behavior
+    # see docs: https://docs.python.org/3/library/pickle.html#object.__getstate
     def __getstate__(self):
         return dict(bucket_name=self.bucket_name, key=self.key, _path=self._path)
 
@@ -74,11 +75,11 @@ class LocalBlob(blobstore.Blob):
         if self.url != src_blob.url:
             shutil.copyfile(src_blob._path, self._path)
 
-    def download(self, path: str):
+    def download(self, target: str):
         if not os.path.isfile(self._path):
             raise blobstore.BlobNotFoundError(f"Could not find {self.url}")
-        if self.url != path:
-            shutil.copyfile(self._path, path)
+        if self._path != target:
+            shutil.copyfile(self._path, target)
 
     def exists(self) -> bool:
         if os.path.isdir(self._path):
--- a/terra_notebook_utils/blobstore/local.py
+++ b/terra_notebook_utils/blobstore/local.py
@@ -41,8 +41,10 @@ class LocalBlob(blobstore.Blob):
         self.key = relpath
         self._path = os.path.join(basepath, relpath)
 
-    # The next two methods customize pickling behavior
-    # see docs: https://docs.python.org/3/library/pickle.html#object.__getstate
+    # The next two methods customize pickling behavior. Some modules such as multiprocessing/ProcessPoolExecutor
+    # require picklable objects. Let's make pickling snappy.
+    # Pickle docs: https://docs.python.org/3/library/pickle.html#object.__getstate
+    # multiprocessing docs: https://docs.python.org/3/library/multiprocessing.html#programming-guidelines
     def __getstate__(self):
         return dict(bucket_name=self.bucket_name, key=self.key, _path=self._path)
 
--- a/terra_notebook_utils/blobstore/local.py
+++ b/terra_notebook_utils/blobstore/local.py
@@ -1,0 +1,117 @@
+import os
+import shutil
+from math import ceil
+from functools import wraps
+from typing import Dict, Generator, Optional
+
+from getm import default_chunk_size
+
+from terra_notebook_utils import blobstore
+
+
+def catch_blob_not_found(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except FileNotFoundError as ex:
+            raise blobstore.BlobNotFoundError(f"Could not find {self.url}") from ex
+    return wrapper
+
+class LocalBlobStore(blobstore.BlobStore):
+    def __init__(self, basepath: str):
+        self.bucket_name = basepath
+
+    def list(self, prefix: str="") -> Generator["LocalBlob", None, None]:
+        root = os.path.join(self.bucket_name, prefix)
+        if root.endswith(os.path.sep):
+            root = root[:-len(os.path.sep)]
+        for (dirpath, dirnames, filenames) in os.walk(root):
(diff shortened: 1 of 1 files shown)


terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+        return LocalBlob(self.bucket_name, key)
+
+class LocalBlob(blobstore.Blob):
+    def __init__(self, basepath: str, relpath: str):
+        assert basepath == os.path.abspath(basepath)
+        self.bucket_name = basepath
+        self.key = relpath
+        self._path = os.path.join(basepath, relpath)
+
+    # The next two methods customize pickling behavior
```
DailyDreaming, line 44: Pickling behavior?  I need more info.  Preferably in this comment, but also in general.

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+        for (dirpath, dirnames, filenames) in os.walk(root):
+            for filename in filenames:
+                relpath = os.path.relpath(os.path.join(dirpath, filename), self.bucket_name)
+                yield LocalBlob(self.bucket_name, relpath)
+
+    def blob(self, key: str) -> "LocalBlob":
+        return LocalBlob(self.bucket_name, key)
+
+class LocalBlob(blobstore.Blob):
+    def __init__(self, basepath: str, relpath: str):
```
DailyDreaming, line 38: Is basepath always the same directory (`pwd`)?  Would it make it simpler to just accept one absolute (or relative) path?

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+    def put(self, data: bytes):
+        with open(self._path, "wb") as fh:
+            fh.write(data)
+
+    @catch_blob_not_found
+    def delete(self):
+        os.remove(self._path)
+
+    @catch_blob_not_found
+    def copy_from(self, src_blob: "LocalBlob"):
```
DailyDreaming, line 69: I would expect this method to be able to copy from any `Blob`.  Is there a reason it's restricted here?

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        Intra-cloud copy
```
DailyDreaming, line 71: This is a non-cloud copy.  :p

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+    def wrapper(self, *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except FileNotFoundError as ex:
+            raise blobstore.BlobNotFoundError(f"Could not find {self.url}") from ex
+    return wrapper
+
+class LocalBlobStore(blobstore.BlobStore):
+    def __init__(self, basepath: str):
+        self.bucket_name = basepath
```
DailyDreaming, line 23: Using `self.bucket_name` for local files, eh?

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+        os.remove(self._path)
+
+    @catch_blob_not_found
+    def copy_from(self, src_blob: "LocalBlob"):
+        """
+        Intra-cloud copy
+        """
+        assert isinstance(src_blob, type(self))
+        if self.url != src_blob.url:
+            shutil.copyfile(src_blob._path, self._path)
```
DailyDreaming, line 75: Is there a reason to avoid the `SameFileError`?  Just curious if this tries to copy an object to itself very often.

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def download(self, path: str):
+        if not os.path.isfile(self._path):
```
DailyDreaming, line 78: Should this also check the `os.path.isdir` edge case here?

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+            raise blobstore.BlobNotFoundError(f"Could not find {self.url}")
+        if self.url != path:
+            shutil.copyfile(self._path, path)
```
DailyDreaming, line 81: `self._path` and `self.url` are the same thing as far as I can tell?  Is there a reason to distinguish them as two different variables in this function?

DailyDreaming (approved):
LGTM.  See comments.

terra_notebook_utils/blobstore/local.py
```diff
@@ -0,0 +1,117 @@
...
+def catch_blob_not_found(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except FileNotFoundError as ex:
+            raise blobstore.BlobNotFoundError(f"Could not find {self.url}") from ex
+    return wrapper
+
+class LocalBlobStore(blobstore.BlobStore):
```
DailyDreaming, line 21: This may also want to optionally accept the `file://` schema.

DailyDreaming (approved):

terra_notebook_utils/blobstore/local.py
xbrianh: I'll add a link to the documentation:
https://docs.python.org/3/library/pickle.html#object.__getstate__

terra_notebook_utils/blobstore/local.py
xbrianh: `copy_from` is for intra-cloud copies, which are schematically different than cross-cloud copies, and often much simpler.

terra_notebook_utils/blobstore/local.py
xbrianh: :)

terra_notebook_utils/blobstore/local.py
xbrianh: The source object needs to be a file.

terra_notebook_utils/blobstore/local.py
xbrianh: I'll use the same variable.

terra_notebook_utils/blobstore/local.py
xbrianh: basepath can be any directory.

This mechanism is more complex than needed, but was pulled from working code so I don't want to make a lot of changes now. I'll consider simplifications in the future.

terra_notebook_utils/blobstore/local.py
xbrianh: This is intended to mimic the cloud blob interface.

terra_notebook_utils/blobstore/local.py
xbrianh: Why bother with SameFileError? Seems easier to just do nothing in this case.

terra_notebook_utils/blobstore/local.py
xbrianh: I prefer not to unless that use case crops up.

terra_notebook_utils/blobstore/local.py
DailyDreaming: Oh, I'm familiar with `pickle` as library and a serialization method.  Just curious why it was accounted for in this code.  Are you doing any pickling?

PR #298, opened 2021-06-07
Introduce url blobstore

xbrianh:
depends on #295 
related to #294

### DIFF
--- a/terra_notebook_utils/blobstore/url.py
+++ b/terra_notebook_utils/blobstore/url.py
@@ -1,4 +1,5 @@
 from functools import wraps
+from typing import Tuple
 
 from requests.exceptions import HTTPError, ConnectionError
 from getm.reader import URLRawReader, URLReaderKeepAlive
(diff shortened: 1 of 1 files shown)


terra_notebook_utils/blobstore/url.py
```diff
@@ -0,0 +1,88 @@
...
+def catch_blob_not_found(func):
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        try:
+            return func(self, *args, **kwargs)
+        except (HTTPError, ConnectionError) as ex:
+            raise blobstore.BlobNotFoundError(f"Could not find {self.url}") from ex
+    return wrapper
+
+class URLBlobStore(blobstore.BlobStore):
```
DailyDreaming, line 25: I would perhaps make `schema` a `list`, as this is inheriting the same `schema = ""` that the local blobstore is, but here that might be confusing.

terra_notebook_utils/blobstore/url.py
```diff
@@ -0,0 +1,88 @@
...
+    def blob(self, url: str) -> "URLBlob":
+        return URLBlob(url)
+
+class URLBlob(blobstore.Blob):
+    def __init__(self, url: str):
+        assert url.startswith("http://") or url.startswith("https://")
+        self.url = self.key = url
```
DailyDreaming, line 32: I would not expect a variable named `self.key` to still contain the `schema`.

DailyDreaming (approved):
LGTM.  See comments.

terra_notebook_utils/blobstore/url.py
xbrianh: will do

terra_notebook_utils/blobstore/url.py
xbrianh: Since http/s URLs are being abstracted as blobs, I don't have a problem with the key containing the schema.

PR #307, opened 2021-06-08
Use new blobstore modules

### LINKED ISSUE #294: Add extensible support for cloud storage

xbrianh:
closes #294

### DIFF
--- a/terra_notebook_utils/cli/drs.py
+++ b/terra_notebook_utils/cli/drs.py
@@ -3,7 +3,7 @@ import json
 import argparse
 from typing import Any, Dict
 
-from terra_notebook_utils import drs, MULTIPART_THRESHOLD
+from terra_notebook_utils import drs
 from terra_notebook_utils.cli import dispatch, Config
 from terra_notebook_utils.drs import DRSResolutionError
 
@@ -60,7 +60,7 @@ def drs_copy_batch(args: argparse.Namespace):
 @drs_cli.command("head", arguments={
     "drs_url": dict(type=str),
     "--bytes": dict(type=int, required=False, default=1, help="Number of bytes to fetch."),
-    "--buffer": dict(type=int, required=False, default=MULTIPART_THRESHOLD,
+    "--buffer": dict(type=int, required=False, default=None,
                      help="Control the buffer size when fetching data."),
     ** workspace_args,
 })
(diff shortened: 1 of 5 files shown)


terra_notebook_utils/drs.py
```diff
@@ -168,40 +175,36 @@ def resolve_drs_for_gs_storage(drs_url: str) -> Tuple[gs.Client, DRSInfo]:
...
     assert drs_url.startswith("drs://"), f'Not a DRS schema: {drs_url}'
     enable_requester_pays(workspace_name, workspace_namespace)
     try:
-        client, info = resolve_drs_for_gs_storage(drs_url)
-        blob = client.bucket(info.bucket_name, user_project=workspace_namespace).blob(info.key)
-        with gscio.Reader(blob, chunk_size=buffer) as handle:
-            the_bytes = handle.read(num_bytes)
-
+        blob = get_drs_blob(drs_url, workspace_namespace)
+        chunk_size = buffer or 2 * num_bytes
```
DailyDreaming, line 205: Is the default buffer size always twice as big as the actual data it's fetching (`2 * num_bytes`)?  Would this still work if it just defaults to the size of the data (`num_bytes`)?

terra_notebook_utils/cli/drs.py
```diff
@@ -60,7 +60,7 @@ def drs_copy_batch(args: argparse.Namespace):
 @drs_cli.command("head", arguments={
     "drs_url": dict(type=str),
     "--bytes": dict(type=int, required=False, default=1, help="Number of bytes to fetch."),
-    "--buffer": dict(type=int, required=False, default=MULTIPART_THRESHOLD,
+    "--buffer": dict(type=int, required=False, default=None,
```
DailyDreaming, line 63: Could the default value (or the description that all data will be loaded without a buffer) be included in the `help`?

DailyDreaming (approved):
LGTM.  See comments.

terra_notebook_utils/cli/drs.py
xbrianh: The size of `buffer` is no longer meaningful.

This should be revisited in a future PR. I will file an issue.

PR #312, opened 2021-06-09
Config to continue or bail when a copy errors

xbrianh:
This adds a configuration parameter to CopyClient to either attempt all copies regardless of failures, or to bail out on the first failure.

depends on #310


DailyDreaming (approved):
LGTM.

PR #313, opened 2021-06-09
Yield part sizes from 'download' and 'copy_from'

xbrianh:
This makes the blob interface a little more challenging to use.
There is a trade-off:
- If 'download' and 'copy_from' do not report the size of parts copied, progress reporting must be embedded into blobstore.blob code, increasing complexity and scattering progress logic across several modules.
-  On the other hand, if 'download' and 'copy_from' DO report part sizes, progress reporting logic can be consolidated into the 'copy_client' module.

On the whole, the second option seems most attractive.

### DIFF
--- a/terra_notebook_utils/blobstore/copy_client.py
+++ b/terra_notebook_utils/blobstore/copy_client.py
@@ -25,7 +25,7 @@ def _download(src_blob: AnyBlob, dst_blob: LocalBlob):
     if dirname:
         os.makedirs(dirname, exist_ok=True)
     # The download methods for each Blob is expected to compute checksums
-    for part_size in src_blob.download(dst_blob.url):
+    for part_size in src_blob.download_iter(dst_blob.url):
         pass
 
 def _copy_intra_cloud(src_blob: AnyBlob, dst_blob: AnyBlob):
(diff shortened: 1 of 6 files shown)


terra_notebook_utils/blobstore/copy_client.py
```diff
@@ -34,7 +35,8 @@ def _copy_intra_cloud(src_blob: AnyBlob, dst_blob: AnyBlob):
     # compute the destination S3Etag on the fly.
     logger.debug(f"Starting intra-cloud {src_blob.url} to {dst_blob.url}")
     assert isinstance(src_blob, type(dst_blob))
-    dst_blob.copy_from(src_blob)  # type: ignore
+    for part_size in dst_blob.copy_from(src_blob):  # type: ignore
```
DailyDreaming, line 38: I liked the simplicity of just calling `download()` or 'copy_from()` as a one-liner earlier.  If there a way to shuttle the `for` loop into those functions and avoid it here?

DailyDreaming (approved):
LGTM.  I think I agree with the design choice.

terra_notebook_utils/blobstore/copy_client.py
xbrianh: Perhaps the current methods should be renamed to `download_iter` and `copy_from_iter`, and convenience one-liners can be introduced as such:
```
def download(self, path):
    for part in self.download_iter(path):
         pass
```

I think I'll do this.

xbrianh: btw, copy_client would need to use the `_iter` versions since it needs to report progress.

PR #322, opened 2021-06-11
Use ordered part iterators and writers

xbrianh:
The unordered logic is more complex and doesn't provide compelling performance. Dump it.
    
Introduce local part writer.


DailyDreaming (approved):
LGTM.

PR #324, opened 2021-06-11
Optionally create subdirs for local blob writes

xbrianh:


terra_notebook_utils/blobstore/local.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, basepath: str, relpath: str):
         self.bucket_name = basepath
         self.key = relpath
         self._path = os.path.join(basepath, relpath)
+        self.makedirs = True  # create subdirectories if needed during write operations
```
DailyDreaming, line 56: What is the use-case to make this configurable (will it ever be 
`False` or set to `False`)?

DailyDreaming (approved):
LGTM.  See comment.

terra_notebook_utils/blobstore/local.py
xbrianh: Depends on the user. 

This behavior (refusing to create subdirectories) is common enough that it seems reasonable to make it configurable.

PR #325, opened 2021-06-11
Refator tar_gz for new blobstore

xbrianh:


terra_notebook_utils/cli/drs.py
```diff
@@ -87,8 +87,11 @@ def drs_head(args: argparse.Namespace):
 
 @drs_cli.command("extract-tar-gz", arguments={
     "drs_url": dict(type=str),
-    "dst_gs_url": dict(type=str, help=("Root of extracted archive. This must be a Google Storage location prefixed"
-                                       "prefixed with 'gs://'")),
+    "dst": dict(type=str,
```
DailyDreaming, line 90: Since this isn't required, is this supposed to be a positional argument?

DailyDreaming (approved):
LGTM.

terra_notebook_utils/cli/drs.py
xbrianh: Yes. If omitted the tarball will be extracted into the workspace bucket.

This may not be the most intuitive behavior. Alternative suggestions are welcome (which will be handled in a future PR).

PR #327, opened 2021-06-14
Organize command modules into commands package

xbrianh:
This should help make CLI code more comprehensible.

PR #329, opened 2021-06-14
Better CLI organization and entrypoint

xbrianh:
- Use an entrypoint command instead of a standalone script to
  define the main CLI binary.
- Introduce dev_scripts/tnu to issue CLI commands in a dev env.
- Rearrange CLI code into sensible submodules.

depends on #326 #327


DailyDreaming (approved):
LGTM.

xbrianh:
Some CLI module reorganization has been backed out due to a failing mock.patch test that I don't wish to investigate right now.

PR #332, opened 2021-06-16
Output json-like data from table list-rows cli

xbrianh:
This makes it easier to work with row IO on the CLI, and produces output more similar to the Terra UI.

For instance this now works:
```
tnu table list-rows | jq
```


DailyDreaming (approved):
LGTM.

PR #336, opened 2021-06-17
Introduce local WDL test using Terra container

xbrianh:
Test is performed against the current master branch on GitHub.

depends on #335

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,6 +5,7 @@ before_script:
   - virtualenv ~/venv
   - source ~/venv/bin/activate
   - pip install -r requirements-dev.txt
+  - apt-get install shellcheck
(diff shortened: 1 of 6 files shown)


.gitlab-ci.yml
```diff
@@ -5,6 +5,7 @@ before_script:
   - virtualenv ~/venv
   - source ~/venv/bin/activate
   - pip install -r requirements-dev.txt
+  - apt-get install shellcheck
```
DailyDreaming: I almost feel like there should be a `requirements-dev.system` file.

DailyDreaming (approved):
LGTM.

.gitlab-ci.yml
xbrianh: That'd be handy!

PR #344, opened 2021-06-25
Avoid writing empty CLI config file

xbrianh:
Other than the pointlessness of writing an empty config file, this prevents intermittent CI/CD false negatives.


DailyDreaming (approved):
LGTM.

PR #345, opened 2021-06-29
Consolidate loggers

xbrianh:
It makes sense for tnu to use a single logger (since it is an
interdependent set of modules)
 - easier to set the log level for the library as a whole
 - easier to reformat log messaging in the future


DailyDreaming (approved):
LGTM.

PR #347, opened 2021-07-01
Update CLI config file data into defaults

xbrianh:
This prevents loss of default configurations that aren't present
in the config file. For instance, when updating TNU with new
CLI configuration values.


DailyDreaming (approved):
LGTM.

PR #348, opened 2021-07-01
Split blobstore progress into separate module

xbrianh:
- Choice of indicator type is no longer globally configured. This
  should be less confusing.
- Prepare the way for additional indicator types, such as IPython
  widgets:
  https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#IntProgress


DailyDreaming (approved):
LGTM.

PR #357, opened 2021-09-24
md5 property for GSBlob

xbrianh:
This is useful for checksum verification for DRS responses containing a signed URL and a known hash.


DailyDreaming (approved):
LGTM.

PR #358, opened 2021-09-24
Settable md5 property for URLBlob

xbrianh:
This is useful checksum verification for Signed URL objects with known md5s.

depends on #357


DailyDreaming (approved):
LGTM.

PR #356, opened 2021-09-22
WIP Add upload to azure functionality

Qi77Qi:
TODO:
Figure out how to use managed identity instead of access key for auth.

Tested on a terra VM

```
>>> drs.copy("drs://jade.datarepo-dev.broadinstitute.org/v1_0c86170e-312d-4b39-a0a4-2a2bfaa24c7a_c0e40912-8b14-43f6-9a2f-b278144d0060", "https://qijlbdgpc4zqdee.blob.core.windows.net/qi-test-container/subdir/jade1")
2021-09-22 05:50:19::INFO  Enabling requester pays for your workspace. This will only take a few seconds...
2021-09-22 05:50:19::WARNING  Failed to init requester pays for workspace qi-monitoring-1218/qi-ws-1: Expected '204', got '401' for 'https://rawls.dsde-prod.broadinstitute.org/api/workspaces/qi-monitoring-1218/qi-ws-1/enableRequesterPaysForLinkedServiceAccounts'. You will not be able to access DRS URIs that interact with requester pays buckets.
2021-09-22 05:50:21::INFO  Request URL: 'https://qijlbdgpc4zqdee.blob.core.windows.net/qi-test-container/subdir/jade1'
2021-09-22 05:50:21::INFO  Request method: 'PUT'
2021-09-22 05:50:21::INFO  Request headers:
2021-09-22 05:50:21::INFO      'x-ms-blob-type': 'REDACTED'
2021-09-22 05:50:21::INFO      'Content-Length': '62043448'
2021-09-22 05:50:21::INFO      'If-None-Match': '*'
2021-09-22 05:50:21::INFO      'x-ms-version': 'REDACTED'
2021-09-22 05:50:21::INFO      'Content-Type': 'application/octet-stream'
2021-09-22 05:50:21::INFO      'Accept': 'application/xml'
2021-09-22 05:50:21::INFO      'User-Agent': 'azsdk-python-storage-blob/12.9.0 Python/3.7.10 (Linux-5.4.104+-x86_64-with-debian-buster-sid)'
2021-09-22 05:50:21::INFO      'x-ms-date': 'REDACTED'
2021-09-22 05:50:21::INFO      'x-ms-client-request-id': '8e96e9a6-1bcd-11ec-b6fa-0242ac120005'
2021-09-22 05:50:21::INFO      'Authorization': 'REDACTED'
2021-09-22 05:50:21::INFO  A body is sent with the request
2021-09-22 05:50:23::INFO  Response status: 201
[...]

PR #359, opened 2021-09-24
Passthrough copies verify md5 sums

xbrianh:
Use md5 checksum verification for passthrough copies.

In the current context of TNU, these are copies from URLBlob to GSBlob. This will soon also cover URLBlob -> AzureBlob, see #356 

depends on #357 #358


DailyDreaming (approved):
LGTM.

Issue #369, opened 2021-11-01
TNU DRS operations fail in Terra project-per-workspace (PPW) workspaces (Urgent)


mbaumann-broad:
**Current Behavior:**
When TNU is run to copy/download DRS data in a project-per-wokspace (PPW) workspace (any workspace created after 9/27/2021) the operation produces the following warning and often fails with the following error:
```
drs.copy("drs://dg.4503/93f98458-e816-4e56-9bea-013dc6c0ea4b", ".")
2021-11-01 02:53:26::INFO  Enabling requester pays for your workspace. This will only take a few seconds...
2021-11-01 02:53:26::WARNING  Failed to init requester pays for workspace terra-f20dfb56/mbaumann_tmp_test_tnu_v0_8_2 20211031: Expected '204', got '404' for 'https://rawls.dsde-prod.broadinstitute.org/api/workspaces/terra-f20dfb56/mbaumann_tmp_test_tnu_v0_8_2%2020211031/enableRequesterPaysForLinkedServiceAccounts'. You will not be able to access DRS URIs that interact with requester pays buckets.
2021-11-01 02:53:27::ERROR  copy failed: 'gs://nih-nhlbi-biodata-catalyst-tutorial-genome-data/GWAS/1kg-genotypes/gds_maf001/ALL.chr8.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.bi_maf001.vcf.bgz.gds' to '/home/jupyter/notebooks/mbaumann_tmp_test_tnu_v0_8_2 20211031/edit/ALL.chr8.phase3_shapeit2_mvncall_integrated_v5a.20130502.genotypes.bi_maf001.vcf.bgz.gds'
Traceback (most recent call last):
  File "/home/jupyter/notebooks/packages/terra_notebook_utils/blobstore/copy_client.py", line 69, in _do_copy
    _download(src_blob, dst_blob, indicator_type)
  File "/home/jupyter/notebooks/packages/terra_notebook_utils/blobstore/copy_client.py", line 27, in _download
    with Indicator.get(indicator_type, dst_blob.url, src_blob.size()) as indicator:
  File "/home/jupyter/notebooks/packages/terra_notebook_utils/blobstore/gs.py", line 162, in size
    return self._get_native_blob().size
  File "/home/jupyter/notebooks/packages/terra_notebook_utils/blobstore/gs.py", line 90, in _get_native_blob
[...]

PR #373, opened 2021-11-15
Distinguish between WORKSPACE_NAMESPACE and GOOGLE_PROJECT

jessebrennan:
To accommodate Terra's Project Per Workspace (PPW) changes, we now need
to distinguish between workspace name and Google projects (they were
previously the same).

This addresses (#369)

### DIFF
--- a/terra_notebook_utils/cli/__init__.py
+++ b/terra_notebook_utils/cli/__init__.py
@@ -13,7 +13,7 @@ from typing import Optional, Tuple
 
 import cli_builder
 
-from terra_notebook_utils import version, WORKSPACE_NAME, WORKSPACE_GOOGLE_PROJECT
+from terra_notebook_utils import version, WORKSPACE_NAME, WORKSPACE_NAMESPACE
 from terra_notebook_utils.blobstore.progress import Indicator
 
 
@@ -45,9 +45,9 @@ class CLIConfig:
     @classmethod
     def resolve(cls, override_workspace: Optional[str], override_namespace: Optional[str]) -> Tuple[str, str]:
         workspace = override_workspace or (cls.info['workspace'] or WORKSPACE_NAME)
-        namespace = override_namespace or (cls.info['workspace_namespace'] or WORKSPACE_GOOGLE_PROJECT)
+        namespace = override_namespace or (cls.info['workspace_namespace'] or WORKSPACE_NAMESPACE)
         if not workspace:
-            raise RuntimeError("This command requies a workspace. Either pass in with `--workspace`,"
+            raise RuntimeError("This command requires a workspace. Either pass in with `--workspace`,"
                                " or configure the CLI (see `tnu config --help`). A default may also be configured by"
                                " setting the `WORKSPACE_NAME` env var")
         if namespace is None:
@@ -56,7 +56,7 @@ class CLIConfig:
         if not namespace:
             raise RuntimeError("This command requies a workspace namespace. Either pass in with `--workspace-namespace`"
                                ", or configure the CLI (see `tnu config --help`). A default may also be"
-                               " configued by setting the `WORKSPACE_GOOGLE_PROJECT` env var")
+                               " configued by setting the `WORKSPACE_NAMESPACE` env var")
         return workspace, namespace
 CLIConfig.load()
 
--- a/terra_notebook_utils/cli/__init__.py
+++ b/terra_notebook_utils/cli/__init__.py
@@ -54,9 +54,10 @@ class CLIConfig:
             from terra_notebook_utils.workspace import get_workspace_namespace
             namespace = get_workspace_namespace(workspace)
         if not namespace:
-            raise RuntimeError("This command requies a workspace namespace. Either pass in with `--workspace-namespace`"
-                               ", or configure the CLI (see `tnu config --help`). A default may also be"
-                               " configued by setting the `WORKSPACE_NAMESPACE` env var")
+            raise RuntimeError("This command requires a workspace namespace. "
+                               "Either pass in with `--workspace-namespace`, "
+                               "or configure the CLI (see `tnu config --help`). A default may also be"
+                               " configured by setting the `WORKSPACE_NAMESPACE` env var")
         return workspace, namespace
 CLIConfig.load()
 
--- a/terra_notebook_utils/cli/commands/vcf.py
+++ b/terra_notebook_utils/cli/commands/vcf.py
@@ -1,8 +1,8 @@
 import json
 import argparse
 
-from terra_notebook_utils import vcf
-from terra_notebook_utils.cli import dispatch, CLIConfig
+from terra_notebook_utils import vcf, WORKSPACE_GOOGLE_PROJECT
+from terra_notebook_utils.cli import dispatch
 from terra_notebook_utils.drs import blob_for_url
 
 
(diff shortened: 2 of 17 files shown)


terra_notebook_utils/cli/__init__.py
```diff
@@ -45,9 +45,9 @@ def write(cls):
     @classmethod
     def resolve(cls, override_workspace: Optional[str], override_namespace: Optional[str]) -> Tuple[str, str]:
         workspace = override_workspace or (cls.info['workspace'] or WORKSPACE_NAME)
-        namespace = override_namespace or (cls.info['workspace_namespace'] or WORKSPACE_GOOGLE_PROJECT)
+        namespace = override_namespace or (cls.info['workspace_namespace'] or WORKSPACE_NAMESPACE)
```
DailyDreaming, line 48: We should probably still account for `WORKSPACE_GOOGLE_PROJECT`:

```
namespace = override_namespace or (cls.info['workspace_namespace'] or WORKSPACE_NAMESPACE or WORKSPACE_GOOGLE_PROJECT)
```

Especially since TNU can be used outside of Terra, and outside of Terra, `WORKSPACE_NAMESPACE` may not exist if you aren't tying yourself to a workspace but still want to pay with a billing account.

terra_notebook_utils/cli/__init__.py
```diff
@@ -45,9 +45,9 @@ def write(cls):
...
         if not workspace:
-            raise RuntimeError("This command requies a workspace. Either pass in with `--workspace`,"
+            raise RuntimeError("This command requires a workspace. Either pass in with `--workspace`,"
```
DailyDreaming, line 50: Thanks, good catch.

DailyDreaming (commented):
Looks good so far.  See comments.

terra_notebook_utils/cli/commands/vcf.py
```diff
@@ -1,27 +1,22 @@
...
-        type=str,
-        default=None,
-        help="Workspace name. If not provided, the configured CLI workspace will be used."
-    ),
-    "--workspace-namespace": dict(
+    "--billing-project": dict(
         type=str,
         required=False,
-        default=CLIConfig.info['workspace_namespace'],
+        default=WORKSPACE_GOOGLE_PROJECT,
```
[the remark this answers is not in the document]
jessebrennan: https://github.com/DataBiosphere/terra-notebook-utils/pull/373/files/b9d4cf5a84eb2df231767c111b7c3cf14d2cfa6a#diff-d236d9cafbc2a61103cba63084cb916010d2008bc873ffb63ac427250bd1076eR9

sort of handles this, but I agree this should be improved.

terra_notebook_utils/drs.py
```diff
@@ -111,7 +111,7 @@ def access(drs_url: str,
         return gs.get_signed_url(bucket=info.bucket_name,
                                  key=info.key,
                                  sa_credentials=info.credentials,
-                                 requester_pays_user_project=workspace_namespace)
+                                 requester_pays_user_project=billing_project)
```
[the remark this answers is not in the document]
jessebrennan: Good question. `enable_requester_pays(workspace_name, workspace_namespace)` does an API request to `rawls` which, from [their docs](https://rawls.dsde-prod.broadinstitute.org/#/workspaces/enableRequesterPaysForLinkedServiceAccounts):

> Enable Requester Pays for the user's linked service accounts in a workspace. Grants the serviceusage.services.use permission in the **project associated to the workspace** to each service account linked in the account linking service (Bond).

In my testing it seems to work as you'd expect; the **project**—associated with the workspace (which we specified with namespace and name)—gets the required permission. Essentially, Terra's API speak the language of `workspace_namespace` and `workspace_name` while Google APIs should receive an actual Google billing project. This is why we use the *actual* Google project below.

terra_notebook_utils/cli/__init__.py
jessebrennan: I agree, we should make it easier to configure a billing project. Currently you can pass the API parameter directly, or set the environment variable(s).

I wanted to refactor parts of how the CLIConfig worked, and make it easier to add more variables, like `WORKSPACE_GOOGLE_PROJECT`. To avoid scope creep, I left it out of this PR, hoping that the current configuration options are good enough for now.

What do you think?

mbaumann-broad: @DailyDreaming, I agree that it would be good to treat configuration and override of all three variables (workspace namespace, workspace name, and google project) consistently. I realize you are working without the benefit of the detailed review I provided to @jessebrennan regarding the PPW changes to these variables and how they are used. The suggestion you provided would not provide the effect you intend.
Also, as you say, it is important that TNU work well outside of Terra, yet in all the cases where the google project is needed (for data access), the workspace_namespace would also be needed for the call to enable requester pays.

@jessebrennan I agree with both your points above. For consistency, CLIConfig should be extended to support all three variables. Conceptually it fits within the scope of this PR, yet there are a lot of code touchpoints in this PR already that function properly as they stand, so I am fine with deferring the CLIConfig change to a subsequent PR.

mbaumann-broad (changes requested):
Thanks @jessebrennan! Very nice work on this!

I reviewed the code and also installed from your branch into a new Terra (PPW) workspace and did some testing there. Overall, it looks really good.

There are a few small change requests, mostly to comments and help text.
Please incorporate these then I'll do a final review and some final testing.

Thank you!

terra_notebook_utils/drs.py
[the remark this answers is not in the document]
jessebrennan: Good point, I added a comment.

terra_notebook_utils/cli/__init__.py
DailyDreaming: @jessebrennan @mikebaumann I was under the impression that workspaces created before a certain date should still have the same behavior and defaults, including the final default of `WORKSPACE_GOOGLE_PROJECT`.  But this default looks like it was never used, as even our test workspace has:

```
WORKSPACE_NAMESPACE=firecloud-cgl
GOOGLE_PROJECT=firecloud-cgl
```

So I suspect `WORKSPACE_GOOGLE_PROJECT` never is and probably never was actually reached as a default.  I believe my misunderstanding was not with the new PPW changes, but that this default was actually needed for the old functionality.  I'm fine with discarding this default now.

terra_notebook_utils/drs.py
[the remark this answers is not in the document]
DailyDreaming: @jessebrennan Thanks for taking the time to test it.  That's satisfactory to me.

DailyDreaming (approved):
LGTM and the controlled access tests pass for me.  Thanks for making the changes.

terra_notebook_utils/cli/__init__.py
mbaumann-broad: Thanks for the clarification/elaboration, @DailyDreaming.
Yes, essentially, my understanding is that `WORKSPACE_GOOGLE_PROJECT` is a TNU-internal name representing the `GOOGLE_PROJECT` environment variable. I don't think there ever was a `WORKSPACE_GOOGLE_PROJECT` environment variable.

mbaumann-broad (approved):
LGTM, too, @jessebrennan!

I reviewed the most recent changes, then installed from this branch to new Cloud Environments in both PPW (recent) and non-PPW (old) workspaces. I tested both the CLI and API for key operations on both, and it all looks good.

Thanks so much for your work on this!
Onward!

terra_notebook_utils/cli/__init__.py
DailyDreaming: @mbaumann-broad Yes, `WORKSPACE_GOOGLE_PROJECT` is not an environment variable, it is a python variable.  The python variable that was set to the env var `GOOGLE_PROJECT` on that python file line could never be reached as a default in Terra.  You could artificially manipulate your environment variables to have the env var `WORKSPACE_NAMESPACE` but not the env var `GOOGLE_PROJECT` but this would never happen in Terra, since both of these are set.  I hope that that clarifies things.

PR #374, opened 2021-11-17
Restore old DRS copy_batch API; separate copy_batch with manifest

jessebrennan:
This is "chained" on top of #369 in order to avoid merge conflicts. I will rebase on master once the other PR lands.

### DIFF
--- a/terra_notebook_utils/drs.py
+++ b/terra_notebook_utils/drs.py
@@ -233,7 +233,7 @@ def _resolve_local_target(filepath: str, info: DRSInfo) -> str:
     # Checking for a file also prevents a batch_copy from overwriting
     # the same file over and over again.
     if os.path.isfile(filepath):
-        raise FileExistsError(filepath)
+        raise FileExistsError(f'Cannot copy {info.name} to {filepath} because file already exists')
     if filepath.endswith(os.path.sep) or os.path.isdir(filepath):
         filename = info.name or info.key.rsplit("/", 1)[-1]
         filepath = os.path.join(os.path.abspath(filepath), filename)
@@ -294,18 +294,21 @@ def copy_to_bucket(drs_uri: str,
         dst_url += f"/{dst_key}"
     copy(drs_uri, dst_url, indicator_type, workspace_name, workspace_namespace)
 
-def copy_batch(manifest: List[Dict[str, str]] = None,
-               drs_urls: Iterable[str] = None,
-               dst_pfx: str = None,
+def copy_batch(manifest: Optional[List[Dict[str, str]]] = None,
+               drs_urls: Optional[Iterable[str]] = None,
+               dst_pfx: Optional[str] = None,
                indicator_type: Indicator = Indicator.notebook_bar if is_notebook() else Indicator.log,
                workspace_name: Optional[str] = WORKSPACE_NAME,
                workspace_namespace: Optional[str] = WORKSPACE_NAMESPACE):
-    assert (manifest is None) != (drs_urls is None), "Specify either 'manifest' or 'drs_urls' and 'dst_pfx'"
-    if manifest:
-        assert dst_pfx is None
+    if (manifest is None) == (drs_urls is None):
+        raise ValueError("Specify either 'manifest' or 'drs_urls' and 'dst_pfx'")
+    elif manifest:
+        if dst_pfx is not None:
+            raise ValueError('dst_pfx not compatible with manifest')
         copy_batch_manifest(manifest, indicator_type, workspace_name, workspace_namespace)
     elif drs_urls:
-        assert dst_pfx is not None, ''
+        if dst_pfx is None:
+            raise ValueError('dst_pfx required with drs_urls')
         copy_batch_urls(drs_urls, dst_pfx, indicator_type, workspace_name, workspace_namespace)
     else:
         assert False
--- a/terra_notebook_utils/drs.py
+++ b/terra_notebook_utils/drs.py
@@ -294,6 +294,34 @@ def copy_to_bucket(drs_uri: str,
         dst_url += f"/{dst_key}"
     copy(drs_uri, dst_url, indicator_type, workspace_name, workspace_namespace)
 
+def copy_batch(manifest: List[Dict[str, str]] = None,
+               drs_urls: Iterable[str] = None,
+               dst_pfx: str = None,
+               indicator_type: Indicator = Indicator.notebook_bar if is_notebook() else Indicator.log,
+               workspace_name: Optional[str] = WORKSPACE_NAME,
+               workspace_namespace: Optional[str] = WORKSPACE_NAMESPACE):
+    assert (manifest is None) != (drs_urls is None), "Specify either 'manifest' or 'drs_urls' and 'dst_pfx'"
+    if manifest:
+        assert dst_pfx is None
+        copy_batch_manifest(manifest, indicator_type, workspace_name, workspace_namespace)
+    elif drs_urls:
+        assert dst_pfx is not None, ''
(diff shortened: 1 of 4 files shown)


DailyDreaming (commented):
@jessebrennan Linting:

```
flake8 terra_notebook_utils tests *.py
terra_notebook_utils/cli/commands/drs.py:95: error: Argument 1 to "copy_batch" has incompatible type "List[Dict[str, Any]]"; expected "Iterable[str]"
Found 1 error in 1 file (checked 53 source files)
```

jessebrennan:
@DailyDreaming the problem is a little more complex than that, but I appreciate you pointing out the error. I'm working on it though, and I'll ping you when I'm ready for your review.

DailyDreaming (commented):
What do you think about copy-batch having the same behavior in both the CLI and the API?

As a user, I can see how it would be easy to assume if `tnu drs copy-batch --manifest=some_manifest.json` works, then `tnu.drs.copy_batch(manifest='some_manifest.json')` would work.  I believe this is also the same throughout the rest of the code base (though correct me if I'm wrong).

Other than that, this looks good to me.

terra_notebook_utils/drs.py
```diff
@@ -302,10 +306,22 @@ def copy_to_bucket(drs_uri: str,
...
 
-def copy_batch(manifest: List[Dict[str, str]],
-               indicator_type: Indicator=Indicator.notebook_bar if is_notebook() else Indicator.log,
-               workspace_name: Optional[str]=WORKSPACE_NAME,
-               workspace_namespace: Optional[str]=WORKSPACE_NAMESPACE):
+def copy_batch(drs_urls: Iterable[str],
+               dst_pfx: str,
+               indicator_type: Indicator = Indicator.notebook_bar if is_notebook() else Indicator.log,
+               workspace_name: Optional[str] = WORKSPACE_NAME,
+               workspace_namespace: Optional[str] = WORKSPACE_GOOGLE_PROJECT):
```
mbaumann-broad, line 313: This:  
`workspace_namespace: Optional[str] = WORKSPACE_GOOGLE_PROJECT):`
must be changed to:  
`workspace_namespace: Optional[str]=WORKSPACE_NAMESPACE):`

mbaumann-broad (changes requested):
Thanks for your work on this @jessebrennan!

It looks good overall, yet there is a small but critical change required.

Please make this change and I will continue my review and testing.

Thank you!

terra_notebook_utils/drs.py
jessebrennan: Whoops, that must have slipped through when I rebased. Nice catch.

terra_notebook_utils/drs.py
```diff
@@ -290,6 +294,34 @@ def copy_to_bucket(drs_uri: str,
...
+               workspace_namespace: Optional[str] = WORKSPACE_NAMESPACE):
+    assert (manifest is None) != (drs_urls is None), "Specify either 'manifest' or 'drs_urls' and 'dst_pfx'"
+    if manifest:
+        assert dst_pfx is None
+        copy_batch_manifest(manifest, indicator_type, workspace_name, workspace_namespace)
+    elif drs_urls:
+        assert dst_pfx is not None, ''
+        copy_batch_urls(drs_urls, dst_pfx, indicator_type, workspace_name, workspace_namespace)
+    else:
+        assert False
```
DailyDreaming, line 311: Could this be upgraded to a proper error with a message?

terra_notebook_utils/drs.py
```diff
@@ -290,6 +294,34 @@ def copy_to_bucket(drs_uri: str,
...
 
+def copy_batch(manifest: List[Dict[str, str]] = None,
+               drs_urls: Iterable[str] = None,
+               dst_pfx: str = None,
+               indicator_type: Indicator = Indicator.notebook_bar if is_notebook() else Indicator.log,
+               workspace_name: Optional[str] = WORKSPACE_NAME,
+               workspace_namespace: Optional[str] = WORKSPACE_NAMESPACE):
+    assert (manifest is None) != (drs_urls is None), "Specify either 'manifest' or 'drs_urls' and 'dst_pfx'"
+    if manifest:
+        assert dst_pfx is None
```
DailyDreaming, line 305: Could a message be added here (preferably containing `dst_pfx`)?

terra_notebook_utils/drs.py
```diff
@@ -290,6 +294,34 @@ def copy_to_bucket(drs_uri: str,
         dst_url += f"/{dst_key}"
     copy(drs_uri, dst_url, indicator_type, workspace_name, workspace_namespace)
 
+def copy_batch(manifest: List[Dict[str, str]] = None,
+               drs_urls: Iterable[str] = None,
+               dst_pfx: str = None,
```
DailyDreaming, line 299: The above three variables (`manifest`, `drs_urls`, and `dst_pfx`) need the `Optional` type.

terra_notebook_utils/drs.py
```diff
@@ -290,6 +294,34 @@ def copy_to_bucket(drs_uri: str,
...
+               indicator_type: Indicator = Indicator.notebook_bar if is_notebook() else Indicator.log,
+               workspace_name: Optional[str] = WORKSPACE_NAME,
+               workspace_namespace: Optional[str] = WORKSPACE_NAMESPACE):
+    assert (manifest is None) != (drs_urls is None), "Specify either 'manifest' or 'drs_urls' and 'dst_pfx'"
+    if manifest:
+        assert dst_pfx is None
+        copy_batch_manifest(manifest, indicator_type, workspace_name, workspace_namespace)
+    elif drs_urls:
+        assert dst_pfx is not None, ''
```
DailyDreaming, line 308: Could a message also be added here?

terra_notebook_utils/drs.py
```diff
@@ -230,6 +230,10 @@ def _resolve_bucket_target(url: str, info: DRSInfo) -> Tuple[str, str]:
     return bucket_name, key
 
 def _resolve_local_target(filepath: str, info: DRSInfo) -> str:
+    # Checking for a file also prevents a batch_copy from overwriting
+    # the same file over and over again.
+    if os.path.isfile(filepath):
+        raise FileExistsError(filepath)
```
DailyDreaming, line 236: Could this error message say something like `Attempting to batch copy {info.name} to {filepath}, but this path would be overwritten, aborting.`

DailyDreaming (changes requested):
Overall, looks good.  Just a few comments, mostly to help with debuggability in the future.

terra_notebook_utils/drs.py
jessebrennan: I changed the other asserts to be more explicit exceptions to make the distinction clear, but this assert will never fire. `assert False` is just a helpful message to the programmer that this code path (under normal operations) should never be reached.

terra_notebook_utils/drs.py
jessebrennan: I reworded and removed `batch` from the message since this applies to all copies, but otherwise :+1:

terra_notebook_utils/drs.py
DailyDreaming: This could be reached if the user specifies either `drs.copy_batch(manifest=[])` or `drs.copy_batch(drs_urls=[])`.  In which case, what would be the best behavior?  A message to the user that nothing was specified and we're exiting?  I don't think it should be a value error or an `assert False` either.

DailyDreaming (changes requested):
Looks good, just one comment.  Thanks for addressing the other asserts.

DailyDreaming (approved):
LGTM.

PR #376, opened 2021-12-02
Reorder copy_batch() parameters for backwards compatibility

jessebrennan:


mbaumann-broad (approved):
Thanks for your work on this, @jessebrennan, @DailyDreaming!

I am happy with the functionality now, much appreciated.

LGTM, although it would still be good to have some information about the manifest in the README.md. Currently it does not contain any information regarding the manifest.

PR #382, opened 2022-03-10
Restore list_bucket()

aofarrel:
Used by the BYOD notebook. This was previously in gs.py but was removed sometime in summer 2021.  I also added a function to strip the gs:// part of the bucket name. If we do not strip that part, list_blobs() will throw a 404 error. I do vaguely recall having to strip the gs:// part before... not sure why it wasn't getting removed in this notebook.

Another option would be to hardcode this into the notebook itself, and that's what I did to test this. But it probably belongs here.


DailyDreaming (approved):
LGTM.

PR #384, opened 2022-03-10
Check if bucket is empty string

aofarrel:
os.environ.get('WORKSPACE_BUCKET') in list_bucket() is resulting in a 404 if the bucket name is not specified when this function is called. Based on the output I have a hunch os.environ.get('WORKSPACE_BUCKET') is returning an empty string.

A possible alternative to this based on get_access_token(), although I'm not sure if it would make a difference:
```
if os.environ.get('WORKSPACE_BUCKET'):
        bucket = os.environ['WORKSPACE_BUCKET']
```


DailyDreaming (changes requested):
This is a no op.  `not bucket` also checks the empty string.

DailyDreaming:
- [ ]

DailyDreaming:
@aofarrel `os.environ.get('WORKSPACE_BUCKET')` returns `None` btw, not an empty string.

PR #386, opened 2022-04-15
sometimes submission parts are not workflows since they lacked some i…

yfarjoun:
…nput...so they do not have a workflowId

This is a tiny PR to protect against that.

I'm not sure how to setup a test for it, but woudl be happy to if someone pointed me in the right direction


DailyDreaming (approved):
This looks good to me.  Thanks for the submission and apologies that this was overlooked for so long.

DailyDreaming:
This seems explanatory enough without adding a test.

PR #389, opened 2022-06-30
Update Iterable import.

DailyDreaming:
Apparently the `typing` library is deprecating the `Iterable` type in python3.10, and the old import is failing our tests.


DailyDreaming:
My bad, this is a firecloud library issue: https://github.com/broadinstitute/fiss/issues/170

PR #392, opened 2022-06-30
Only request an access URL if using access, copy, or copy-batch.

DailyDreaming:
Getting the signed URL requires an additional round-trip to the DRS server, so we shouldn't ask Martha to fetch one unless we need to.


mbaumann-broad (approved):
LGTM, thank you @DailyDreaming!

DailyDreaming:
Thanks for the review @mbaumann-broad !

PR #393, opened 2022-06-30
Get rid of unnecessary info call.

DailyDreaming:
I believe that this is all that's needed.


mbaumann-broad (approved):
Looks right to me! Thank you!

DailyDreaming:
Thanks again!

PR #397, opened 2022-11-23
[ID-323] Allow tnu to use either Martha or DRSHub to resolve DRS URIs 

Shakespeared:
This PR is to make the DRS resolver configurable, so DRS URIs can be resolved using either Martha or DRShub. You will be able to change the service called by updating the new env var `DRS_RESOLVER`

We are working on replacing Martha with DRSHub, so there will be a gradual cutover and eventually Martha will be deprecated. Behaviors should be the same between Martha and DRSHub so you shouldn't notice any difference when you cut over. We're still perf testing DRSHub so you don't need to flip the switch yet, someone will let you know when we're ready to swap. Please let me know if you have any questions about this!

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,3 +1,7 @@
+# Changes for v0.11.0 (2022-11-09)
+Added support for calling DRSHub for DRS resolution
+Made DRS Resolver configurable by setting `DRS_RESOLVER` to `martha` or `drshub`
+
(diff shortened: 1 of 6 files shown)


Shakespeared:
closing this because I originally forked the repo, I am making a replacement PR without forking

PR #398, opened 2022-11-23
[ID-323] Allow tnu to use either Martha or DRSHub to resolve DRS URIs 

Shakespeared:
This PR is to make the DRS resolver configurable, so DRS URIs can be resolved using either Martha or DRShub. You will be able to change the service called by updating the new env var DRS_RESOLVER

We are working on replacing Martha with DRSHub, so there will be a gradual cutover and eventually Martha will be deprecated. Behaviors should be the same between Martha and DRSHub so you shouldn't notice any difference when you cut over. We're still perf testing DRSHub so you don't need to flip the switch yet, someone will let you know when we're ready to swap. Please let me know if you have any questions about this!

### DIFF
--- a/terra_notebook_utils/__init__.py
+++ b/terra_notebook_utils/__init__.py
@@ -5,6 +5,7 @@ WORKSPACE_NAMESPACE = os.environ.get('WORKSPACE_NAMESPACE')  # This env var is s
 WORKSPACE_GOOGLE_PROJECT = os.environ.get('GOOGLE_PROJECT')  # This env var is set in Terra Cloud Environments
 TERRA_DEPLOYMENT_ENV = os.environ.get('TERRA_DEPLOYMENT_ENV', 'prod')
 MARTHA_URL_VERSION = os.environ.get('MARTHA_URL_VERSION', 'martha_v3')
+DRS_RESOLVER = os.environ.get('DRS_RESOLVER', 'martha')  # values for DRS_RESOLVER are either 'drshub' or 'martha'
 
 if not WORKSPACE_GOOGLE_PROJECT:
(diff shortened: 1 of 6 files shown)


terra_notebook_utils/__init__.py
```diff
@@ -5,6 +5,7 @@
 WORKSPACE_GOOGLE_PROJECT = os.environ.get('GOOGLE_PROJECT')  # This env var is set in Terra Cloud Environments
 TERRA_DEPLOYMENT_ENV = os.environ.get('TERRA_DEPLOYMENT_ENV', 'prod')
 MARTHA_URL_VERSION = os.environ.get('MARTHA_URL_VERSION', 'martha_v3')
+DRS_RESOLVER = os.environ.get('DRS_RESOLVER', 'martha')  # values for DRS_RESOLVER are either 'drshub' or 'martha'
```
DailyDreaming, line 8: Should this default to `'drshub'` now?

DailyDreaming (commented):
LGTM, just one comment.

terra_notebook_utils/__init__.py
Shakespeared: Since we haven't fully perf tested DRSHub yet, I think it should stay as Martha for now. However as soon as DRSHub is ready for traffic it should definitely default to DRSHub. Do you want me to make a ticket on my board to switch that when it's ready?

DailyDreaming: Yes please, thanks!

DailyDreaming (approved):
LGTM.

terra_notebook_utils/__init__.py
Shakespeared: Cool, I'll do that right now, then it won't get lost in the shuffle!

Issue #399, opened 2023-01-15
Enable TNU to use Azure auth for DRS operations


mbaumann-broad:
# Objective
Support TNU DRS operations when TNU is running in a Terra Azure IA Cloud Environment.  
When running in a Terra Azure IA Cloud Environment, TNU must get/use an Azure access token (via the `DefaultAzureCredential`) when communicating with Terra backend services (i.e., martha/terra-drs-hub, rawls).  
Support for all TNU operations when running on GCS must continue to work as it has.

# Scope
- This specific change is limited to the TNU DRS functionality, which is needed very soon for a Terra Azure milestone.
- This change does not address other TNU functionality, such as `table`, `workflows`, etc. The implementation of these for Terra Azure may be substantially different than the current implementation for Terra GCS and is therefore out of scope.

# Background
Terra backend services now accept three types of bearer tokens:  
- Google access token (as before). 
- Azure access token. 
- Microsoft B2C access token.  

The proposed algorithm to successfully obtain an appropriate access token for use with Terra backend services is:
1.  Identify the context in which TNU is running. Proposed values include:
  - Terra Azure
  - Terra GCS
  - Unknown (some other context, such as on a local system)
 2. Get an access token based on the execution context
   2.1. if the execution context is Terra Azure, get an Azure access token. 
   2.2. if the execution context is Terra GCS, get a Google access token (as before). 
   2.3. if the execution context is unknown, try getting a GCS access token, if that fails try getting an Azure token, if that fails, then exit with an error.

# Acceptance Criteria
- All TNU DRS operations work successfully in Terra Azure IA Cloud Environments (especially Jupyter CEs)
- The TNU CI/CD test environment is enhanced to represent Terra Azure Cloud Environments.
- All TNU operations on Terra GCS Cloud Environments continue to work as before this change.
- A TNU version with these changes is released and available in PyPi.

Issue #400, opened 2023-01-17
Only enable requester pays when necessary


mbaumann-broad:
# Objective
Only perform the operation to enable requester pays when it is currently necessary.  
The only case where it is currently necessary is when accessing AnVIL data hosted by Gen3.  
This case can be identified by DRS URIs starting with "drs://dg.ANV0".  
In all other cases, enabling requester pays adds unnecessary delays and, in some cases, as when a Terra user has an external identity linkage to Kids First, spurious errors.

# Background
Accessing DRS data hosted by Gen3 in requester pays buckets using Google cloud-native access APIs required that the Gen3 user client service account be granted the `serviceusage.services.use` permission on the Terra user's Google project.

This privileged operation is performed by the Rawls API [enableRequesterPaysForLinkedServiceAccounts](https://rawls.dsde-prod.broadinstitute.org/#/workspaces/enableRequesterPaysForLinkedServiceAccounts).

Originally, this was needed for Terra DRS access to Gen3-hosted data for both Biodata Catalyst and AnVIL.
Terra DRS access to Gen3-hosted Biodata Catalyst data has since changed to using signed URLs, and enabling requester pays is no longer necessary for Biodata Catalyst.
Terra DRS access to Gen3-hosted data for NCI CRDC and Kids First has never been necessary.

When AnVIL data is fully hosted in Terra Data Repo (TDR), enabling requester pays will no longer be needed at all and should be disabled completely.

# Acceptance Criteria
- terra-notebook-utils only enables requester pays when DRS operations are performed on one or more DRS URIs starting with "drs://dg.ANV0"
 - A version of terra-notebook-utils with this functionality has been released and is available from PyPi.

PR #401, opened 2023-01-17
Enable TNU to use Azure auth for DRS operations #399

mbaumann-broad:
For a description of the change, see: #399 

**Status**
- All intended functionality implemented and working as expected
- Automated tests updated and passing
- README updated
- Completed final review
- Ready to merge and release!

### DIFF
--- a/terra_notebook_utils/__init__.py
+++ b/terra_notebook_utils/__init__.py
@@ -1,4 +1,6 @@
 import os
+from dataclasses import dataclass
+from enum import Enum
 
 WORKSPACE_NAME = os.environ.get('WORKSPACE_NAME', None)
 WORKSPACE_NAMESPACE = os.environ.get('WORKSPACE_NAMESPACE')  # This env var is set in Terra Cloud Environments
@@ -26,3 +28,20 @@ if DRS_RESOLVER == "martha":
     DRS_RESOLVER_URL = MARTHA_URL
 else:
     DRS_RESOLVER_URL = DRSHUB_URL
+
+
+class ExecutionEnvironment(Enum):
+    TERRA_WORKSPACE = 1,  # Executing in a Terra Workspace (on any supported platform)
+    OTHER = 2  # Executing outside of a Terra Workspace (e.g., local system)
+
+
+class ExecutionPlatform(Enum):
+    AZURE = 1,  # Executing in an Azure compute environment
+    GOOGLE = 2,  # Executing in a Google compute environment
+    UNKNOWN = 3  # Execution platform not identified (e.g., local system)
+
+
+@dataclass
+class ExecutionContext:
+    execution_environment: ExecutionEnvironment
+    execution_platform: ExecutionPlatform
--- a/terra_notebook_utils/__init__.py
+++ b/terra_notebook_utils/__init__.py
@@ -31,14 +31,14 @@ else:
 
 
 class ExecutionEnvironment(Enum):
-    TERRA_WORKSPACE = 1,  # Executing in a Terra Workspace (on any supported platform)
-    OTHER = 2  # Executing outside of a Terra Workspace (e.g., local system)
+    TERRA_WORKSPACE = "TERRA_WORKSPACE",  # Executing in a Terra Workspace (on any supported platform)
+    OTHER = "OTHER"  # Executing outside a Terra Workspace (e.g., local system)
 
 
 class ExecutionPlatform(Enum):
-    AZURE = 1,  # Executing in an Azure compute environment
-    GOOGLE = 2,  # Executing in a Google compute environment
-    UNKNOWN = 3  # Execution platform not identified (e.g., local system)
+    AZURE = "AZURE",  # Executing in an Azure compute environment
+    GOOGLE = "GOOGLE",  # Executing in a Google compute environment
+    UNKNOWN = "UNKNOWN"  # Execution platform not identified (e.g., local system)
 
 
 @dataclass
--- a/terra_notebook_utils/azure_auth.py
+++ b/terra_notebook_utils/azure_auth.py
@@ -23,7 +23,7 @@ from terra_notebook_utils.logger import logger
 _AZURE_CREDENTIAL: Optional[DefaultAzureCredential] = None
 
 
-def _set_azure_identity_logging_level(level) -> None:
+def _set_azure_identity_logging_level(level: int) -> None:
     """ Set the logging level for modules participating the Azure default credential flow """
     import azure.identity
     logging.getLogger(azure.identity._credentials.environment.__name__).setLevel(level)
--- a/terra_notebook_utils/azure_auth.py
+++ b/terra_notebook_utils/azure_auth.py
@@ -1,13 +1,12 @@
 """
-Microsoft Azure specific code.
-Currently limited to Azure auth.
+Microsoft Azure identity/auth support.
 
 See:
 https://azuresdkdocs.blob.core.windows.net/$web/python/azure-identity/1.12.0/index.html
 https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python
 https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_credentials/default.py
 """
-
+import logging
 import os
 from typing import Optional
 
@@ -22,6 +21,19 @@ from terra_notebook_utils.logger import logger
 # See: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/
 #              azure-identity/azure/identity/_credentials/default.py
 _AZURE_CREDENTIAL: Optional[DefaultAzureCredential] = None
+
+
+def _set_azure_identity_logging_level(level) -> None:
+    """ Set the logging level for modules participating the Azure default credential flow """
+    import azure.identity
+    logging.getLogger(azure.identity._credentials.environment.__name__).setLevel(level)
+    logging.getLogger(azure.identity._credentials.managed_identity.__name__).setLevel(level)
+    logging.getLogger(azure.identity._credentials.chained.__name__).setLevel(level)
+    logging.getLogger("azure.core.pipeline.policies.http_logging_policy").setLevel(level)
+
+
+# Suppress extraneous azure-identity INFO level logging
+_set_azure_identity_logging_level(logging.WARNING)
 
 
 def _get_default_credential() -> DefaultAzureCredential:
--- a/terra_notebook_utils/azure_auth.py
+++ b/terra_notebook_utils/azure_auth.py
@@ -1,0 +1,58 @@
+"""
+Microsoft Azure specific code.
+Currently limited to Azure auth.
+
+See:
+https://azuresdkdocs.blob.core.windows.net/$web/python/azure-identity/1.12.0/index.html
+https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python
+https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_credentials/default.py
+"""
+
+import os
+from typing import Optional
+
+from azure.identity import DefaultAzureCredential
+from terra_notebook_utils.logger import logger
+
+
+# Single instance of DefaultAzureCredential that initialized lazily.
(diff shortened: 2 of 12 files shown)


terra_notebook_utils/drs.py
```diff
@@ -46,7 +47,7 @@ def enable_requester_pays(workspace_name: Optional[str]=WORKSPACE_NAME,
     rawls_url = (f"https://rawls.dsde-{TERRA_DEPLOYMENT_ENV}.broadinstitute.org/api/workspaces/"
                  f"{workspace_namespace}/{encoded_workspace}/enableRequesterPaysForLinkedServiceAccounts")
     logger.info("Enabling requester pays for your workspace. This will only take a few seconds...")
-    access_token = gs.get_access_token()
+    access_token = get_terra_access_token()
```
rtitle, line 50: Thinking -- should we just avoid the `enable_requester_pays` entirely when the execution context is Azure? This is really a GCP-specific function..

rtitle (commented):
Overall looks good. I like the tagging of `execution_platform` and `execution_context`. 

Were you able to test in a real Azure workspace notebook against an Azure DRS URL? It looks like there is still some GCP-specific logic in `drs.py` (such as checking requester-pays, and trying to resolve the file a GCS object if the signed url doesn't work). But hopefully the auth piece is enough to simply make it work for Azure signed URLs.

terra_notebook_utils/utils.py
```diff
@@ -63,3 +66,22 @@ def is_notebook() -> bool:
...
+    Other/improved means may be identified in the future.
+    """
+    execution_environment = ExecutionEnvironment.OTHER
+    execution_platform = ExecutionPlatform.UNKNOWN
+    workspace_bucket = os.environ.get('WORKSPACE_BUCKET', None)
+    if workspace_bucket:
+        execution_environment = ExecutionEnvironment.TERRA_WORKSPACE
+        if workspace_bucket.startswith("gs://"):
+            execution_platform = ExecutionPlatform.GOOGLE
+        elif workspace_bucket.startswith("https://"):
```
nmalfroy, line 85: I don't think we can depend on just using https.  Maybe we could do something like
```suggestion
        elif re.search("^https://.*\.(dfs|blob)\.core\.windows\.net$", workspace_bucket):
```

(assuming you have `import re` above)

terra_notebook_utils/drs.py
mbaumann-broad: Yes, I think that makes sense. Even for Terra GCP, the only remaining case where enabling requester pays is necessary is AnVIL data hosted on Gen3, and even that is going away later this year, as described here #400

terra_notebook_utils/azure_auth.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Note: It would not need to be instantiated this way, as
+    # no exception is raised even if Azure credentials are not configured.
+    :return: Reference to instance of DefaultAzureCredential
+    """
+
+    # Should a more sophisticated Singleton pattern be used instead?
+    global _AZURE_CREDENTIAL
+    if not _AZURE_CREDENTIAL:
+        _AZURE_CREDENTIAL = DefaultAzureCredential()
```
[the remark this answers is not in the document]
mbaumann-broad: I would very much like to verify the `DefaultAzureCredential` is valid, yet I didn't see a way to do that. It only has one public member/method and that is `get_token`.
https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python

The `DefaultAzureCredential` is instantiated lazily the first time a Azure access token is requested, and retained for future use for the life of the program.
Hopefully this is okay. I didn't find any information about the expected lifecycle or intended threadsafety of this object.
Any knowledge/experience on this topic for this object would be greatly appreciated!

terra_notebook_utils/drs.py
```diff
@@ -11,13 +11,14 @@
     WORKSPACE_GOOGLE_PROJECT
 from terra_notebook_utils import workspace, gs, tar_gz, TERRA_DEPLOYMENT_ENV, _GS_SCHEMA
 from terra_notebook_utils.utils import is_notebook
-from terra_notebook_utils.http import http
+from terra_notebook_utils.http_utils import http
```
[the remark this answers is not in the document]
mbaumann-broad: I hope not. I didn't want to make the change, yet python was confused by there being both a package and a module named `http`. The use of the `http` package may have been introduced by these Azure auth changes, as apparently it had not been a conflict before.
I addressed the conflict by renaming the module `http.py` to `http_utils.py` which did resolve the conflict and I haven't yet seen indication of breakage.
Yet I am still somewhat concerned about this.

terra_notebook_utils/terra_auth.py
```diff
@@ -0,0 +1,70 @@
...
+    def _identify_valid_access_token() -> str:
+        """
+        Try to obtain an auth bearer tokens suitable for use with Terra backend services
+        from the Terra supported auth providers. First try Google, then try Azure.
+        Return the first successfully obtained token, otherwise raise AuthenticationError.
+
+        :return: auth bearer token suitable for use with Terra backend services
+        :raises: AuthenticationError
+        """
+        try:
```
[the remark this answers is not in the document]
mbaumann-broad: The current code tries to identify the platform, and in cases where we can, we get an auth token for the platform to communicate with Terra backend services.

Yet, based on the current mechanism of identifying the platform, we cannot identify the platform in all cases. In this event, we try both Google and Azure and use what works.
When running in a Terra Cloud Environment, we should be able to identify the platform reliably.
The case where we cannot is when TNU is running external to Terra, perhaps in a cloud VM or local/institutional system. This is not a common case, yet it works today, and I would like to retain that capability.

I welcome ideas for how we can determine the platform (and appropriate default platform auth system) reliably!

terra_notebook_utils/utils.py
mbaumann-broad: Based on recent Slack messages, it sounds like the env vars to be used for Terra Azure Cloud Environments, and what they represent, is in flux.
We can revisit this when that settles out.
And, yes, happy to use a more specific regex pattern match if we have a pattern that we are confident will always be correct.

nmalfroy: Ah ok gotcha

terra_notebook_utils/terra_auth.py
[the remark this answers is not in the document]
nmalfroy: Ah ok.  I thought it was only within Terra that it ran.  I think it's fair to separate from this ticket but I do think there should be some way to explicitly set the cloud platform for TNU.  But given that, your approach makes sense

terra_notebook_utils/azure_auth.py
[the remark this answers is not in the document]
rtitle: I think the approach seems reasonable -- `DefaultAzureCredential` is a singleton instantiated lazily, and `get_token` is called on demand to get an access token.

terra_notebook_utils/drs.py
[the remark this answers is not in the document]
nmalfroy: I guess I was worried that someone else might be using the `terra_notebook_utils.http` module.  in that case, their stuff would break I think.

Where was the conflict occurring?  If it's in this file, for instance, you could do something like:
`from terra_notebook_utils.http import http as http_tnu`

terra_notebook_utils/drs.py
mbaumann-broad: @rtitle, this is now addressed/resolved in PR #402.

For example, in Terra Azure, when attempting to access a DRS URI entails requester pays:
```
jupyter@vm07858848f:~$ tnu drs copy drs://dg.ANV0/1b1ee6fc-6560-4b08-9c44-36d46bf4daf1 .
RequesterPaysNotSupported: Requester pays data access is not supported on the Azure platform. Cannot access: drs://dg.ANV0/1b1ee6fc-6560-4b08-9c44-36d46bf4daf1
```
The new code checks if requester pays is required for any of the given DRS URIs before proceeding to enable requester pays. If requester pays is required, and the platform is Azure, the error above is generated.

As a backup, the function to enable requester pays now includes an an assertion that the platform is not Azure:
```
def enable_requester_pays(workspace_name: Optional[str]=WORKSPACE_NAME,
                          workspace_namespace: Optional[str]=WORKSPACE_NAMESPACE):

    assert get_execution_context().execution_platform != ExecutionPlatform.AZURE, \
        "Requester pays data access is not supported on the Terra Azure platform."
```

terra_notebook_utils/drs.py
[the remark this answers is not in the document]
nmalfroy: ah ok.  I really appreciate you digging into it!

nmalfroy (approved):
looks good to me!

terra_notebook_utils/utils.py
mbaumann-broad: Opened this ticket and noted the above, to be addressed in the (near) future:  
[Reliable identification of primary cloud #403](https://github.com/DataBiosphere/terra-notebook-utils/issues/403)

terra_notebook_utils/terra_auth.py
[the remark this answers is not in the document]
mbaumann-broad: Opened this ticket and noted the above, to be addressed in the (near) future:  
[Reliable identification of primary cloud #403](https://github.com/DataBiosphere/terra-notebook-utils/issues/403)

terra_notebook_utils/azure_auth.py
```diff
@@ -0,0 +1,70 @@
...
+# Single instance of DefaultAzureCredential that initialized lazily.
+# The instance is treated as threadsafe and reusable.
+# The Azure documentation is silent on thread safety.
+# Based on scanning the code, it appears to be threadsafe.
+# See: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/
+#              azure-identity/azure/identity/_credentials/default.py
+_AZURE_CREDENTIAL: Optional[DefaultAzureCredential] = None
+
+
+def _set_azure_identity_logging_level(level) -> None:
```
DailyDreaming, line 26: `level` could use Typing here.

terra_notebook_utils/terra_auth.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Provides auth bearer tokens suitable for use with Terra backend services.
+    """
+    def __init__(self):
+        self.execution_context = get_execution_context()
+
+    @staticmethod
+    def _identify_valid_access_token() -> str:
+        """
+        Try to obtain an auth bearer tokens suitable for use with Terra backend services
```
DailyDreaming, line 26: Nit: `an auth bearer tokens`

terra_notebook_utils/__init__.py
```diff
@@ -26,3 +28,20 @@
...
+
+class ExecutionEnvironment(Enum):
+    TERRA_WORKSPACE = 1,  # Executing in a Terra Workspace (on any supported platform)
+    OTHER = 2  # Executing outside of a Terra Workspace (e.g., local system)
+
+
+class ExecutionPlatform(Enum):
+    AZURE = 1,  # Executing in an Azure compute environment
+    GOOGLE = 2,  # Executing in a Google compute environment
+    UNKNOWN = 3  # Execution platform not identified (e.g., local system)
```
DailyDreaming, line 41: Nit:  In a debugger, these are kind of magic numbers.  It might be clearer to simply use strings (i.e. "google") instead of numbers.  This also has the benefit that a user can print "ExecutionContext.execution_platform" and see "google" or "azure" instead of "2" or "1".

DailyDreaming (approved):
Overall, this looks good and I'd be fine merging it as is.  I included some comments (only nits though).

nmalfroy (approved):
Thanks for making all of the extra changes!

terra_notebook_utils/azure_auth.py
mbaumann-broad: Thank you, level is now typed as `int`

terra_notebook_utils/terra_auth.py
mbaumann-broad: Corrected, thanks for the careful review!

terra_notebook_utils/__init__.py
mbaumann-broad: Good idea! Enum values are now (uppercase) string values.

PR #405, opened 2023-05-02
ID-376 Use DRSHub in Dev

tlangs:
Since we're retiring Martha, we need to make TNU talk to DRSHub instead. This is the first step in that process, making TNU talk to DRSHub in Dev. We'll be running a suite of tests in Dev to make sure things look good before making TNU talk to DRSHub in Prod.


Shakespeared (approved):

DailyDreaming (approved):
LGTM!

PR #406, opened 2023-07-26
ID-423 Let Terra Choose DRS Endpoint

tlangs:
Basically, we don't want to release TNU every time an endpoint migrates or per-environment, so lets make it configurable.


Ghost-in-a-Jar (approved):

tlangs:
The previous way we were doing this was overly complicated. Lets not do that.

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #420, opened 2026-03-13
CTM-381:Always send the workspace google project to the DRSHub request body

LizBaldo:
Jira Ticket: https://broadworkbench.atlassian.net/browse/CTM-381

All TDR clients via DRSHub should now pass the Google Project ID of the requester so the DRSHub resolution can then be billed appropriately if requester pay is enabled.


jgainerdewar (approved):

lucymcnatt (approved):

aednichols (approved):
LGTM, similar to Cromwell a "why?" comment with story number could be very useful.

